PR #8, opened 2020-09-25
Update tokens contract to handle multi tx unstaking

eonwarped:
Instead of subtracting all stake on the unstake event, only subtract the tokens that will be released next. On unstake processing, continue to subtract the next amount to be released.

Update total staked tokens to match this computation.

Update delegation validation to account for new stake balance.


bt-cryptomancer (approved):
Changes look good, thanks!

PR #9, opened 2020-09-25
Add mining contract

eonwarped:
The mining contract achieves parity with scotbot, allowing lotteries to run at scheduled intervals, and issuing tokens directly to winners.

It allows for specifying up to two mining tokens, which must be stakable, and a multiplier for each. The issuer of the token bring mined is the only one that can create or update a mining pool for their token, and only one pool can be set up per configuration of (mined token, mining tokens).

On creation or on multiplier update, the contract will compute the mining power for each staked user.

Also whenever stake changes happen the mining contract will be called to update the mining power for that user.

To verify the lottery results, one needs to to either replay with a node or recompute the winning number from the seed, and then run the results by ordering the participants by mining power descending.


bt-cryptomancer (changes requested):
Overall this looks pretty good, but before merging it can you please take care of the following:

1. Can you update the PR by merging hive-engine into your branch, so the tokens contract changes from your previous PR don't show up as new additions in this one?

2. Add a starting Hive block for the checkPendingLotteries virtual transaction. Should be some time after the proposed release date of Oct. 6. Also add a "contract doesn't exist" safeguard check (see code a few lines below the virtual transaction push).

3. For the checkPendingLotteries action, can you add a maximum limit to the number of such lotteries that can be checked per block? See tick action of botcontroller contract for such an example.  Be careful that this doesn't introduce any non-determinism (i.e. order of DB data results should always be the same).

4. I'm a bit concerned about how expensive an operation initMiningPower could potentially be, since it loops over all the balance records for a particular token (which could number in the thousands for popular tokens). What are your thoughts on the performance of this? I wonder if it would be possible to add a max limit to how many balances could be processed per block, so that large numbers are broken up among multiple blocks? Any required state information that would have to be maintained between blocks could be stored temporarily in a separate table. Without doing something like this, I'm worried it might work OK for now but wouldn't scale well as the Engine platform grows and more users join.

bt-cryptomancer:
Also, just another optimization thought:  it looks like initMiningPower will create a miningPower record for every account that has a token balance, even if that account doesn't have anything staked and thus isn't mining anything.  You could use less storage space and speed up processing a bit by excluding accounts that have 0 staked or delegated from the balances query. 
 Then if those accounts stake or get a delegation later on, they'll just have a new record inserted.

eonwarped:
Should have resolved all issues and updated tests.

bt-cryptomancer (approved):
Changes look good, thanks!

PR #11, opened 2020-10-02
allow hive engine owner to create and update pool for utility token

eonwarped:


contracts/mining.js
```diff
@@ -223,7 +223,7 @@ actions.updatePool = async (payload) => {
   // eslint-disable-next-line no-template-curly-in-string
   const utilityTokenBalance = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: "'${CONSTANTS.UTILITY_TOKEN_SYMBOL}$'" });
 
-  const authorized = api.BigNumber(poolUpdateFee).lte(0)
+  const authorized = api.BigNumber(poolUpdateFee).lte(0) || api.sender === api.owner
```
bt-cryptomancer, line 226: Here you are skipping the balance check for the fee if api.owner is calling the contract action, but fee is still deducted further down.  Maybe api.owner should be excluded from paying the fee as it's a special case?

contracts/mining.js
```diff
@@ -291,7 +292,7 @@ actions.createPool = async (payload) => {
   // eslint-disable-next-line no-template-curly-in-string
   const utilityTokenBalance = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: "'${CONSTANTS.UTILITY_TOKEN_SYMBOL}$'" });
 
-  const authorizedCreation = api.BigNumber(poolCreationFee).lte(0)
+  const authorizedCreation = api.BigNumber(poolCreationFee).lte(0) || api.sender === api.owner
```
bt-cryptomancer, line 295: Here you are skipping the balance check for the fee if api.owner is calling the contract action, but fee is still deducted further down.  Maybe api.owner should be excluded from paying the fee as it's a special case?

bt-cryptomancer (commented):
Mostly looks good, just want clarification regarding fee handling for api.owner.

contracts/mining.js
eonwarped: Fixed

bt-cryptomancer (approved):
changes look good, thanks

PR #13, opened 2020-10-07
Fix unstake, check available balance for unstaking. 

eonwarped:
Prevents unstaking stake amounts that are in pending unstake balance, applicable for multi tx transactions. Add test to catch error case. Delegations already applies validation here.


bt-cryptomancer (approved):
changes look good

PR #15, opened 2020-10-10
separate active status flip in setActive action. do not allow power t…

eonwarped:
…o be negative.


bt-cryptomancer (approved):
changes look good

PR #17, opened 2020-10-22
Airdrops Smart Contract

ali-h:
adding airdrops smart contract, airdrops test file, virtual transaction every block for it.

### DIFF
--- a/contracts/airdrops.js
+++ b/contracts/airdrops.js
@@ -102,7 +102,7 @@ const transferIsSuccesfull = (result, action, from, to, symbol, quantity) => {
     && el.event === action
     && el.data.from === from
     && el.data.to === to
-    && el.data.quantity === quantity
+    && api.BigNumber(el.data.quantity).eq(quantity)
     && el.data.symbol === symbol) !== undefined) {
     return true;
   }
@@ -144,13 +144,13 @@ actions.initAirdrop = async (payload) => {
           to: CONTRACT_NAME, symbol, quantity: airdrop.quantity,
         });
 
-        if (transferIsSuccesfull(tokenTransfer, 'transferToContract', api.sender, 'airdrops', symbol, airdrop.quantity)) {
+        if (transferIsSuccesfull(tokenTransfer, 'transferToContract', api.sender, CONTRACT_NAME, symbol, airdrop.quantity)) {
           // deduct fee from sender's utility token balance
           const feeTransfer = await api.executeSmartContract('tokens', 'transfer', {
             to: 'null', symbol: UTILITY_TOKEN_SYMBOL, quantity: airdrop.fee, isSignedWithActiveKey,
           });
 
-          if (transferIsSuccesfull(feeTransfer, 'transfer', api.sender, 'null', symbol, airdrop.fee)) {
+          if (transferIsSuccesfull(feeTransfer, 'transfer', api.sender, 'null', UTILITY_TOKEN_SYMBOL, airdrop.fee)) {
             await api.db.insert('pendingAirdrops', {
               id: api.transactionId,
               symbol,
--- a/contracts/airdrops.js
+++ b/contracts/airdrops.js
@@ -1,7 +1,10 @@
+const UTILITY_TOKEN_SYMBOL = "'${CONSTANTS.UTILITY_TOKEN_SYMBOL}$'"
+const CONTRACT_NAME = 'airdrops'
+
 actions.createSSC = async () => {
   const tableExists = await api.db.tableExists('pendingAirdrops');
   if (tableExists === false) {
-    await api.db.createTable('pendingAirdrops', ['txId', 'symbol', 'sender']);
+    await api.db.createTable('pendingAirdrops', ['txId', 'symbol']);
     await api.db.createTable('params');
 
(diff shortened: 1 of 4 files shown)


contracts/airdrops.js
```diff
@@ -0,0 +1,244 @@
...
+
+  // check if staking is enabled
+  if (type === 'stake' && api.assert(token.stakingEnabled === true, 'staking not enabled')) {
+    return true;
+  }
+
+  return false;
+};
+
+const transferIsSuccesfull = (result, action, from, to, symbol, quantity) => {
```
eonwarped, line 98: typo:  rename to `transferIsSuccessful`

contracts/airdrops.js
```diff
@@ -0,0 +1,244 @@
...
+
+  if (api.assert(isSignedWithActiveKey === true, 'you must use a custom_json signed with your active key')
+    && api.assert(symbol && typeof symbol === 'string'
+      && type && typeof type === 'string'
+      && list && typeof list === 'object' && list.length, 'invalid params')
+    && api.assert(type === 'transfer' || type === 'stake', 'invalid type')) {
+    const token = await api.db.findOneInTable('tokens', 'tokens', { symbol });
+
+    // get api.sender's utility and airdrop token balances
+    const { balance: utilityTokenBalance } = { ...await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: UTILITY_TOKEN_SYMBOL }) };
```
eonwarped, line 128: Clearer if this is unwrapped

const utilityTokenBalance = (await api.db.findOne... ).balance;

But you probably want to store the await in some temporary variable and check if it is present. (can combine below with e.g.

const utilityTokenBalance = await api.db.findOne...
if (...  utilityTokenBalance && utilityTokenBalance.balance && api.BigNumber(utilitytokenBalance.balance).... )

contracts/airdrops.js
```diff
@@ -0,0 +1,244 @@
...
+        quantity,
+      });
+
+      // add this quantity to the total quantity of tokens to airdrop
+      airdrop.quantity = api.BigNumber(airdrop.quantity).plus(quantity);
+    }
+  }
+
+  // calculate total fee
+  airdrop.fee = api.BigNumber(params.feePerTransaction).times(airdrop.list.length);
```
eonwarped, line 75: Probably an edge case, but do you want to round the fee to the right precision?

contracts/airdrops.js
```diff
@@ -0,0 +1,244 @@
...
+    await api.db.update('pendingAirdrops', airdrop);
+  } else {
+    // if no other transactions are remaining, delete airdrop
+    await api.db.remove('pendingAirdrops', airdrop);
+  }
+
+  api.emit('airdropDistribution', {
+    airdropId,
+    list: completedDrops,
+  });
```
eonwarped, line 225: Perhaps you don't need the full completedDrops listed in this event. Just the airdrop ID and how many were processed would be enough. Also you will want to modify account history server later to trail these events and have these show up in the history.

contracts/airdrops.js
```diff
@@ -0,0 +1,244 @@
...
+    symbol,
+    type,
+  } = airdrop;
+
+  let airdropIsPending = true;
+  let count = 0;
+
+  const completedDrops = [];
+
+  while (airdropIsPending) {
```
eonwarped, line 185: can shorten to  

while (count < maxTransactionsPerBlock && list[count] !== undefined) {
..
count += 1
}

eonwarped (commented):
Generally looks good. Just a few comments.

Also one question, do we want to restrict the list size of the airdrop? I suppose it's already gated by the max possible custom json TX though.

contracts/airdrops.js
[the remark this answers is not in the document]
ali-h: good catch, fixed 👍

contracts/airdrops.js
```diff
@@ -0,0 +1,244 @@
...
+  airdrop.quantity = '0';
+  airdrop.isValid = false;
+
+  // loop through list and validate
+  for (let i = 0; i < list.length; i += 1) {
+    const { 0: to, 1: quantity } = list[i];
+
+    if (to && api.isValidAccountName(to)
+      && quantity && !api.BigNumber(quantity).isNaN()
+      && api.BigNumber(quantity).gt(0) && api.BigNumber(quantity).dp() <= precision) {
```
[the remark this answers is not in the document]
ali-h: I didn't get that?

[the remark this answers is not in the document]
eonwarped: Since you will report it as invalid if any entry is not good, you can exit as soon as you detect a problem. Also optional: report entry that failed (can emit a log or just api assert)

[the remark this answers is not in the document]
ali-h: ah, yes. it should not loop through whole list after finding error in the first object.

contracts/airdrops.js
ali-h: hmm, sure. i just didn't give it that thought since its managed by hive-engine team only but a good check indeed.

eonwarped (approved):
Changes look good.

PR #18, opened 2020-10-31
Hive mining nft

eonwarped:
Implements nft into mining.

Uses nft "type" property distinguish different NFTs. Mining config specifies power settings on a per type basis.

Up to four mining properties can be specified. Each of these can configure whether it is accumulated with + or * as the op.

After combining all properties of all owned NFTs, the full power is calculated by * all of the property totals.

NFTs do not have staking, so instead we use delegation to determine whether an NFT is active for mining. Delegating to the mining contract constitutes applying the nft to yourself, and delegating to others constitutes applying the nft to others.

Note that properties can be negative as well. If the op results in negative total power or NaN, note that it will result in 0 power for the account.

### DIFF
--- a/contracts/mining.js
+++ b/contracts/mining.js
@@ -80,9 +80,10 @@ const findAndProcessAll = async (contractName, table, query, callback) => {
   }
 };
 
-async function validateTokenMiners(tokenMiners) {
+async function validateTokenMiners(tokenMiners, nftTokenMiner) {
   if (!api.assert(tokenMiners && Array.isArray(tokenMiners), 'tokenMiners invalid')) return false;
-  if (!api.assert(tokenMiners.length >= 1 && tokenMiners.length <= 2,
+  if (!api.assert((tokenMiners.length >= 1 && tokenMiners.length <= 2)
+      || (nftTokenMiner && tokenMiners.length === 0),
     'only 1 or 2 tokenMiners allowed')) return false;
   const tokenMinerSymbols = new Set();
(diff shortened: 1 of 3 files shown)


contracts/mining.js
```diff
@@ -489,3 +718,30 @@ actions.handleStakeChange = async (payload) => {
...
+  if (api.assert(callingContractInfo && callingContractInfo.name === 'nft',
+    'must be called from nft contract')) {
+    await findAndProcessAll('mining', 'nftTokenPools', { symbol }, async (tokenPool) => {
+      const pool = await api.db.findOne('pools', { id: tokenPool.id });
+      if (pool.updating.inProgress
+          && pool.updating.tokenIndex === pool.tokenMiners.length
+          && pool.updating.lastId < nft._id) {
+        return;
+      }
+      const typeProperties = pool.nftTokenMiner.typeMap[nft.properties.type];
```
bt-cryptomancer, line 735: The "type" property of an NFT instance could be changed to anything at any time.  What happens if it's not something in your typeMap when it gets delegated?

bt-cryptomancer (commented):
Code mostly looks good, just have a few comments I've left at specific code lines which I'd like you to respond to before approving this.

contracts/mining.js
eonwarped: added type check. should be sufficient here.

bt-cryptomancer (approved):
changes look good

PR #22, opened 2020-11-08
Distribution Contract

donchate:
A token distribution contract intended for use in decentralized governance. Collects tokens to their specified threshold, then distributes to a list of recipients by percentage share. Payments occur on first deposit that exceeds a token's threshold or if triggered manually by the owner.

Features:
- Payout threshold per token
- Up to 40 recipients that can be users or other smart contracts
- Update thresholds/recipients at any time
- Activate/Deactivate
- Flush function for owner/creator to distribute tokens that have not reached threshold


contracts/distribution.js
```diff
@@ -0,0 +1,301 @@
...
+
+  await api.db.update('params', params);
+};
+
+async function processBatch(batch, symbol, isFlush = false) {
+  let balance;
+  let balanceInd;
+  let payout;
+
+  for (let i = 0; i < batch.tokenBalances.length; i += 1) {
```
eonwarped, line 48: Suggestion, have balance store the actual object instead of just the quantity, and you can use it below without storing balanceInd also.

This can be shortened to

const balance = batch.tokenBalances.find(b => b.symbol === symbol);

contracts/distribution.js
```diff
@@ -0,0 +1,301 @@
...
+    const curBalance = batch.tokenBalances[i];
+    if (curBalance.symbol === symbol) {
+      balanceInd = i;
+      balance = curBalance.quantity;
+      break;
+    }
+  }
+
+  for (let i = 0; i < batch.tokenMinPayout.length; i += 1) {
```
eonwarped, line 57: const payout = batch.tokenMinPayout.find(p => p.symbol === symbol);

contracts/distribution.js
```diff
@@ -0,0 +1,301 @@
...
+      break;
+    }
+  }
+
+  if (balance !== undefined && payout !== undefined
+    && (api.BigNumber(balance).gt(api.BigNumber(payout)) || isFlush === true)) {
+    // pay out token balance to recipients by configured share percentage
+    for (let i = 0; i < batch.tokenRecipients.length; i += 1) {
+      const recipient = batch.tokenRecipients[i];
+      const recipientShare = api.BigNumber(balance).multipliedBy(recipient.pct / 100).toFixed(3);
```
eonwarped, line 70: instead of dividing by 100 here, I would use BigNumber division after.

contracts/distribution.js
```diff
@@ -0,0 +1,301 @@
...
+      if (await api.transferTokens(recipient.account, symbol, recipientShare, recipient.type)) {
```
eonwarped, line 71: You probably want to check the return value of transferTokens for success.

And we will want to determine what the protocol is on any failures. Should it freeze its state, or should it just fire and forget, and rely on external monitoring to determine if this contract is not behaving properly. One thing that is difficult here is that you dont' have a mechanism of an all-or-nothing rollback mechanism for DB. @bt-cryptomancer , what do you think?

test/distribution.js
```diff
@@ -0,0 +1,824 @@
...
+      };
+
+      await send(blockchain.PLUGIN_NAME, 'MASTER', { action: blockchain.PLUGIN_ACTIONS.PRODUCE_NEW_BLOCK_SYNC, payload: block });
+
+      let res = await database1.find({
+        contract: 'distribution',
+        table: 'batches'
+      });
+  
+      assert.ok(!res, 'invalid distribution created');
```
eonwarped, line 291: i would check error messages so you know the intended validation fired

eonwarped: same comment for other validation checks

contracts/distribution.js
```diff
@@ -0,0 +1,301 @@
...
+        // eslint-disable-next-line no-param-reassign, max-len
+        batch.tokenBalances[balanceInd].quantity = api.BigNumber(batch.tokenBalances[balanceInd].quantity).minus(recipientShare);
+        await api.db.update('batches', batch);
```
bt-cryptomancer, line 74: It's not necessary to update the DB on every pass through this for loop.  It's more efficient to do one single update at the end when the loop is complete.

contracts/distribution.js
bt-cryptomancer: It's very unlikely that a transfer will fail, but it's still good practice to do a sanity check on the result to verify success (see other contracts for examples of how to do this).  In the event of a failure, I would record the error with an api.assert but otherwise allow the processing to continue, rather than stop in some halfway complete state.

contracts/distribution.js
eonwarped: Ah yes. And this is related to the previous comment about storing the balance object. (It will be by reference so it will modify the existing batch object). You would be able to keep updating that and then do the db update at the end of the loop.

contracts/distribution.js
donchate: If we encounter a failed transfer, we can keep the quantity in the batch's balance and thereby reattempt distributing it next time it runs. I have to update my test which always expects the batch to get fully emptied on a payment run, but that seems like the most preferable option. Recipients could be updated later and this leftover balance would eventually distribute.

contracts/distribution.js
donchate: Thanks, much more elegant.

contracts/distribution.js
donchate: Resolved by storing balance reference object.

contracts/distribution.js
bt-cryptomancer: Okay that sounds fine.

contracts/distribution.js
```diff
@@ -0,0 +1,287 @@
...
+  }
+};
+
+actions.deposit = async (payload) => {
+  const {
+    id, symbol, quantity, isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(isSignedWithActiveKey === true, 'you must use a custom_json signed with your active key')
+    && !api.assert(quantity && api.BigNumber(quantity).dp() <= 3 && api.BigNumber(quantity).gt(0), 'invalid quantity')) {
```
eonwarped, line 248: this should be OR ing the asserts?

eonwarped: should have a test to cover these asserts

contracts/distribution.js
```diff
@@ -0,0 +1,287 @@
...
+      // update token balances
+      if (dist.tokenBalances) {
+        let hasBalance = false;
+        for (let i = 0; i < dist.tokenBalances.length; i += 1) {
+          if (dist.tokenBalances[i].symbol === symbol) {
+            dist.tokenBalances[i].quantity += quantity;
+            hasBalance = true;
+            break;
+          }
+        }
```
eonwarped, line 268: optional: change to use find()

eonwarped (approved):
looks good after addressing comments

contracts/distribution.js
donchate: Nice catch and added to tests as well.

contracts/distribution.js
donchate: Still not sure if using find() makes it more readable or not, I do believe however its a slower operation. I have left this alone for now.

eonwarped: find is definitely more readable (no manual breaking and assigning values), and for the case where so few elements are involved, it does not matter.  (I also am searching around and I don't actually think it is slower depending on the engine)

eonwarped (approved):

PR #23, opened 2020-11-13
Add burn mechanism to nft mining

eonwarped:
Properties now can take in a burnChange parameter that allows for burning tokens to change the property for a given fee. This is a proportional fee, e.g. if the burnChange property is 0.1 and the user wants to change a type's prperty by 10, the fee to do so will be 0.1*10 = 1.


bt-cryptomancer (approved):
changes look good

PR #24, opened 2020-11-14
Claimdrops Smart Contract

ali-h:
Claimdrops Smart Contract, can be used to create coin offering with custom prices. creates a pool which has a expiration time upto 90 days (variable) and maximum claims (variable). anyone or selected users can claim tokens for the defined price in Hive Pegged token (similar to market). there can be a claim limit for everyone or individuals in a list which can go up than the pool itself so its first to fill the pool. the price collected, and pool (if remaining) after the expiration is sent to a account or contract defined by owner.

Actions:
* create: creates a claimdrop
fee 50 BEE and 0.1*maxClaims is burned.
* claim: can claim tokens from an active claimdrop
price in SWAP.HIVE is sent to the owner, it is calculated by quantity parameter. price*quantity (price defined in the claimdrop)

### DIFF
--- a/contracts/claimdrops.js
+++ b/contracts/claimdrops.js
@@ -1,7 +1,9 @@
-/* eslint-disable no-await-in-loop */
 /* global actions, api */
 
 const UTILITY_TOKEN_SYMBOL = 'BEE';
+const UTILITY_TOKEN_PRECISION = 8;
+const HIVE_PEGGED_SYMBOL = 'SWAP.HIVE';
+const HIVE_PEGGED_SYMBOL_PRECISION = 8;
 const CONTRACT_NAME = 'claimdrops';
 
 actions.createSSC = async () => {
@@ -11,8 +13,8 @@ actions.createSSC = async () => {
     await api.db.createTable('params');
 
     const params = {};
-    params.creationFee = '1000';
-    params.updateFee = '300';
+    params.creationFee = '50';
+    params.feePerClaim = '0.1';
     await api.db.insert('params', params);
   }
 };
@@ -21,7 +23,7 @@ actions.updateParams = async (payload) => {
   if (api.assert(api.sender === api.owner, 'not authorized')) {
     const {
       creationFee,
-      updateFee,
+      feePerClaim,
     } = payload;
 
     const params = await api.db.findOne('params', {});
@@ -30,27 +32,16 @@ actions.updateParams = async (payload) => {
       if (!api.assert(typeof creationFee === 'string' && !api.BigNumber(creationFee).isNaN() && api.BigNumber(creationFee).gte(0), 'invalid creationFee')) return;
       params.creationFee = creationFee;
     }
-    if (updateFee) {
-      if (!api.assert(typeof updateFee === 'string' && !api.BigNumber(updateFee).isNaN() && api.BigNumber(updateFee).gte(0), 'invalid updateFee')) return;
-      params.updateFee = updateFee;
+    if (feePerClaim) {
+      if (!api.assert(typeof feePerClaim === 'string' && !api.BigNumber(feePerClaim).isNaN() && api.BigNumber(feePerClaim).gte(0), 'invalid feePerClaim')) return;
+      params.feePerClaim = feePerClaim;
     }
 
     await api.db.update('params', params);
   }
 };
 
-const hasValidType = (token, type) => {
-  if (type === 'transfer') {
-    return true;
-  }
-
-  // check if staking is enabled
-  if (type === 'stake' && api.assert(token.stakingEnabled === true, 'staking not enabled')) {
-    return true;
-  }
-
-  return false;
-};
+const hasValidPrecision = (value, precision) => (api.BigNumber(value).dp() <= precision);
 
 const transferIsSuccessful = (result, action, from, to, symbol, quantity) => {
   if (result.errors === undefined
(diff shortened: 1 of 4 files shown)


donchate:
A nice option would be to distribute unclaimed pool amount to those who claimed.

contracts/claimdrops.js
```diff
@@ -0,0 +1,346 @@
...
+  const parsedList = [];
+  for (let i = 0; i < list.length; i += 1) {
+    const { 0: account, 1: limit } = list[i];
+
+    // account & limit validation
+    if (api.assert(account, `list[${i}]: account name cannot be undefined`)
+      && api.assert(api.isValidAccountName(account), `list[${i}]: invalid account name`)
+      && api.assert(limit, `list[${i}]: limit cannot be undefined`)
+      && api.assert(!api.BigNumber(limit).isNaN(), `list[${i}]: invalid limit`)
+      && api.assert(api.BigNumber(limit).gt(0), `list[${i}]: limit must be positive`)
```
bt-cryptomancer, line 93: Can you do a max size check for limit?  Can be something like asserting that api.BigNumber(limit).lte(Number.MAX_SAFE_INTEGER) and also use the BigNumber isFinite function - https://mikemcl.github.io/bignumber.js/#isF

contracts/claimdrops.js
```diff
@@ -0,0 +1,346 @@
...
+    list,
+    limit,
+    isSignedWithActiveKey,
+  } = payload;
+
+  if (api.assert(isSignedWithActiveKey === true, 'you must use a custom_json signed with your active key')
+    && api.assert(symbol && typeof symbol === 'string'
+      && price && typeof price === 'string' && !api.BigNumber(price).isNaN()
+      && pool && typeof pool === 'string' && !api.BigNumber(pool).isNaN()
+      && maxClaims && Number.isInteger(maxClaims)
```
bt-cryptomancer, line 136: Can you do a max size check for maxClaims?  Should make sure it is less than Number.MAX_SAFE_INTEGER

contracts/claimdrops.js
```diff
@@ -0,0 +1,346 @@
...
+        };
+
+        // add list or limit to final claimdrop object
+        if (list) {
+          const parsedList = validateList(list, token.precision);
+          if (parsedList) {
+            claimdrop.list = parsedList;
+          } else return;
+        } else if (limit) {
+          if (api.assert(api.BigNumber(limit).gt(0), 'limit must be positive')
```
bt-cryptomancer, line 198: Can you do a max size check for limit? Can be something like asserting that api.BigNumber(limit).lte(Number.MAX_SAFE_INTEGER) and also use the BigNumber isFinite function - https://mikemcl.github.io/bignumber.js/#isF

bt-cryptomancer (changes requested):
This mostly looks good.  I think it would be good to have a bit stricter validation on max size limits, see comments at indicated code sections.

I'd also like to avoid adding another virtual transaction to Block.js, as we already have several of those and I don't want to add any more unless absolutely necessary.  Let's have checkExpiredClaimdrops be called first whenever someone uses the claim action.  And it would be good to have an expire action that only the claimdrop owner can use, so the owner can manually trigger expiration when desired, instead of waiting for the next claim action to do it.

bt-cryptomancer:
> A nice option to add would be to distribute unclaimed pool amount to those who claimed.

Agreed, I don't think this is necessary for the first version of the contract, but would be a cool feature to add in a version 2 release.

ali-h:
> A nice option to add would be to distribute unclaimed pool amount to those who claimed.

a nice feature, this can surely be added in the next version as an optional choice for the owner.

ali-h:
> This mostly looks good. I think it would be good to have a bit stricter validation on max size limits, see comments at indicated code sections.
> 
> I'd also like to avoid adding another virtual transaction to Block.js, as we already have several of those and I don't want to add any more unless absolutely necessary. Let's have checkExpiredClaimdrops be called first whenever someone uses the claim action. And it would be good to have an expire action that only the claimdrop owner can use, so the owner can manually trigger expiration when desired, instead of waiting for the next claim action to do it.

I've added stricter validations so its less than MAX_SAFE_INTEGER but is `isFinite` necessary? if the value is less than safe integer its finite right.

Yes, initially I was going to add expireClaimdrops in the claim action, but thought it might get the processing slow? but its probably better than adding virtual transaction every block for it. so not searching for all expired claimdrops, now when a claim action is called, if the requested claimdrop has reached its expiration time it gets deleted (the expire action triggers).

I've modified `checkExpiredClaimdrops`  to be `expire` which can be called by the sender or owner set by the sender of a claimdrop, this gives the option to expire a claimdrop manually anytime without depending on claim action.

I've added sender and owner both to the claimdrop object because maybe if the owner is a contract it can not call expire event.

bt-cryptomancer (changes requested):
> I've added stricter validations so its less than MAX_SAFE_INTEGER but is isFinite necessary? if the value is less than safe integer its finite right.

eonwarped started using isFinite in his contract code, which I thought was a nice idea, but you're right that's a redundant check since you're comparing with Number.MAX_SAFE_INTEGER anyway.  So this is fine.

> Yes, initially I was going to add expireClaimdrops in the claim action, but thought it might get the processing slow? but its probably better than adding virtual transaction every block for it. so not searching for all expired claimdrops, now when a claim action is called, if the requested claimdrop has reached its expiration time it gets deleted (the expire action triggers).

Expiring a claimdrop is a pretty cheap operation, so I'd still like to see a claim operation expire all old claimdrops, not just the one being claimed.  The market contract works similarly; when a market order is placed, it removes all old orders which have already expired.

All other changes look fine.

bt-cryptomancer (changes requested):
The way you've got the removeExpiredClaimdrops function coded now introduces the possibility of non-deterministic behavior.  Although this is extremely unlikely with this kind of contract, it's still something that needs to be handled properly.  The problem is that api.db.find implicitly limits query results to 1000 when you don't explicitly provide the limit parameter.  Thus if there are more than 1000 expired claimdrops, this code won't work as intended (and without using indexes on the query, the order of query results is not well defined).

I see 2 options to fix this problem:

1.  Add back in the maxExpiresPerBlock setting that you originally had, as well as the use of the _id index.

OR

2. Wrap the query in a while loop to ensure the case of > 1000 query results is handled OK.  For an example of this technique, see the actions.checkPendingUndelegations function in nft.js or upgradeUserSchema function in botcontroller.js

bt-cryptomancer:
Of the two options I suggested to resolve the query issue, if using number 2 I think it's still important to make sure an index is used for sorting.  We probably should do an audit of the other contracts at some point and ensure indexes are always being used, or else change the core node software to always add the _id index if one is not specified.

ali-h:
changes made. I did think about this but got up with the likely scenario. but yup contracts can be abused if there are no checks for things like this. I've added similar loop to `checkPendingUndelegations` in nft contract where loop continues if there are more than 1000 objects in the db with sorting by _id.

bt-cryptomancer (approved):
changes look good

PR #25, opened 2020-11-17
do not add undelegating nft balances on pool power update computation

eonwarped:
Fix test.

Note that there was an issue with an existing test where we issued a delegate op for something that was undelegating, and that did not throw an error (may be something to fix validation for, but it doesn't hurt anything for the time being).


bt-cryptomancer (approved):
changes look good

PR #26, opened 2020-11-17
Fix cannot make NFT sell order with fee of 0

donchate:
``if fee`` evaluates to false when its a integer of 0.

I believe typeof should be sufficient to determine whether its a `number` or `undefined`, which I think was the original intent.


bt-cryptomancer (approved):
changes look good

PR #28, opened 2020-12-24
Fix over-subtraction of mining power during unstake

eonwarped:
There are two issues:  The first is that on the last token release (no stake change), it does an additional stake deduction on the incorrect amount. The second is that there is a mismatch between when stake is deducted and mining power is deducted. This change consolidates the stake changes in a single place to prevent the confusion and correct the issue.


bt-cryptomancer (approved):
changes look good

PR #29, opened 2021-01-01
Add logging to table hash computation

eonwarped:
This is to be able to debug all hash computation changes for a node to be able to cross check with other nodes.


bt-cryptomancer (changes requested):
I agree this is useful for debugging, but will make log output quite noisy.  Can you add a setting to config.json to control whether this is enabled?  Can be a boolean flag called verboseLogging or something similar.

Also PR should be against the hive-engine branch, not master (we generally no longer use master, which is for Steem Engine).

eonwarped:
Oops right, will redo

PR #30, opened 2021-01-07
Mongo DB Transactions, Entity Caching, Replayability and Hash comparison

eonwarped:
1. Add Mongo DB Transactions for safer replays. Each block opens up a Mongo Transaction, and either data from the whole block commits or none at all.
2. Entity cache introduced for certain cases that abort transactions due to no primary keys
3. Hash comparison functionality to the main node allows for consistency checks for backups, in preparation for witness. (Can be disabled with config)
4. Witness contract prepared for deployment. Witness scheduling is set for a far way block, to be moved closer right before cutting new release / deployment


bt-cryptomancer (approved):
changes look good to me

PR #31, opened 2021-01-15
fix replay with blocklog

eonwarped:
Allow the blockchain plugin to regenerate virtual transactions from sidechain blocklog replay. There are some sidechain blocks that do not have transactions which should also be processed on replay mode, before it got skipped.


bt-cryptomancer (approved):
changes look good

PR #32, opened 2021-01-20
Update witness contract, scripts

eonwarped:
* witness action script added
* more debugging on sync hash script
* allow loading partial plugins (e.g. if you only want to run rpc server)
* fix witness contract so that pending unstake is ignored (stake already properly accounts for it)


bt-cryptomancer (approved):
changes look good

PR #34, opened 2021-01-21
Fix witness round hash, change reward amount to match inflation contract

eonwarped:
Missed rounds were happening because some round computations were done before the block was ready. Added a delay / retry mechanism to accomodate that.

Fix witness reward amount to match inflation.


bt-cryptomancer (approved):
changes look good

PR #35, opened 2021-01-23
Add block lookahead behavior to fetch blocks far in advance to speed up replay

eonwarped:
Among all listed nodes, it will send 3 QPS (configurable) at a time to each node to fetch blocks, and allows for utilizing faster nodes more efficiently (as they free up their requests more quickly). But broken nodes will slow down the whole process, so this will need some monitoring (todo-- automatically kick out repeatedly failing nodes)

### DIFF
--- a/config.json
+++ b/config.json
@@ -9,9 +9,7 @@
     "streamNodes": [
         "https://api.deathwing.me",
         "https://rpc.ecency.com",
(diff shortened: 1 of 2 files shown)


bt-cryptomancer (approved):
changes look good

PR #36, opened 2021-01-24
disable cache on reads if no session (e.g. RPC / P2P)

eonwarped:
Also makes sure no operations flush caches (e.g. reads) if there is no session, which could be problematic.


eonwarped:
Linted and tested locally

bt-cryptomancer (approved):
changes look good

PR #38, opened 2021-01-27
Fix P2P stall issue in certain cases

eonwarped:
await on verify round handler to properly handle errors, and add null check for last proposed round


bt-cryptomancer (approved):
changes look good

PR #42, opened 2021-01-29
small random weight selection fix to witness contract, witness contract cleanup

eonwarped:
fix random selection of witnesses, it is off by lowest precision amount. See https://stackoverflow.com/questions/1527803/generating-random-whole-numbers-in-javascript-in-a-specific-range for formula

This is a difference between the two following formulas:

floor( rnd * ( TOTAL - ACC )) + ACC + eps

vs

floor( rnd * (TOTAL  - ACC - eps) + ACC + eps )

where (eps  = smallest positive token amount, ACC is top witnesses approval weight sum,  TOTAL is total witnesses approval weight sum)

Also changes rounding modes to use floor instead of round as per https://mikemcl.github.io/bignumber.js/#toFix

We should probably audit other usages of random and fix accordingly. It's a very minimal impact in general though.

### DIFF
--- a/contracts/witnesses.js
+++ b/contracts/witnesses.js
@@ -540,13 +540,12 @@ const manageWitnessesSchedule = async () => {
           // calculate a random weight if not done yet
           if (schedule.length >= NB_TOP_WITNESSES
             && randomWeight === null) {
-            const min = api.BigNumber(accWeight)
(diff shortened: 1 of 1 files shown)


bt-cryptomancer (approved):
changes look good

PR #44, opened 2021-02-05
change number of top witnesses, add params, use sidechain block as sw…

eonwarped:
change number of top witnesses, add params, use sidechain block as switch block


eonwarped:
Todo: change test hive ref block numbers to be consistent for the witness contract update operation

eonwarped:
More todos from chat: change min signatures required to 9 and change p2p plugin to handle transition. 

Witness contract itself should also have a gate for when to apply the change (when schedule size >= 11)

eonwarped:
Adjustments made. Ready to review.

PR #45, opened 2021-02-06
Witness contract to allow domains in p2p endpoint

eonwarped:


scot-nitrous:
Dropping request, p2p will remain ip.

PR #46, opened 2021-02-08
Automated market making pools for token exchange

donchate:
This contract provides an additional decentralized exchange (DEX) for hive engine tokens using automated market making with a liquidity protocol. The pricing model uses the constant product formula (```x * y = k```) with no fees taken from traders. Pools are created and funded in a manner similar to other well-known implementations and swaps are intended to be executed in as simple a manner as possible (i.e. buy X tokens for Y, buy Y tokens for X).

An internal oracle to the hive engine order book provides a sanity check when initializing new pool liquidity, and ensures the pool starts with a realistic market price.

Every swap is checked for maximum slippage to prevent a large order relative to available liquidity pool from dramatically affecting the price.

Each pool collects some statistics that can be used on various dashboards such as; total volume traded per pair, current price, current liquidity.

Pre-merge docs https://github.com/donchate/steemsmartcontracts-wiki/blob/marketpools/Market-Pools.md

### DIFF
--- a/contracts/marketpools.js
+++ b/contracts/marketpools.js
@@ -327,7 +327,7 @@ actions.swapTokens = async (payload) => {
   } else if (tradeType === 'exactOutput') {
     const tokenAmountAdjusted = api.BigNumber(getAmountIn(tokenAmount, liquidityIn, liquidityOut));
     if (!tokenAmountAdjusted.isFinite()) return;
-    senderBaseFunded = senderBase && api.BigNumber(senderBase.balance).gte(tokenAmountAdjusted);
+    senderBaseFunded = senderBase && api.BigNumber(senderBase.balance).gte(tokenAmountAdjusted.toFixed(tokenIn.precision, api.BigNumber.ROUND_DOWN));
     tokenPairDelta = tokenSymbol === baseSymbol ? [api.BigNumber(tokenAmount).times(-1), tokenAmountAdjusted] : [tokenAmountAdjusted, api.BigNumber(tokenAmount).times(-1)];
     tokenQuantity = { in: tokenAmountAdjusted, out: tokenAmount };
     if (!api.assert(api.BigNumber(tokenQuantity.out).dp() <= tokenOut.precision, 'symbolOut precision mismatch')) return;
--- a/contracts/marketpools.js
+++ b/contracts/marketpools.js
@@ -27,7 +27,9 @@ actions.updateParams = async (payload) => {
 
 function getAmountIn(amountOut, liquidityIn, liquidityOut) {
   if (!api.assert(api.BigNumber(amountOut).gt(0), 'insufficient output amount')
-    || !api.assert(api.BigNumber(liquidityIn).gt(0) && api.BigNumber(liquidityOut).gt(0), 'insufficient liquidity')) return false;
+    || !api.assert(api.BigNumber(liquidityIn).gt(0)
+      && api.BigNumber(liquidityOut).gt(0)
+      && api.BigNumber(amountOut).lt(liquidityOut), 'insufficient liquidity')) return false;
   const num = api.BigNumber(liquidityIn).times(amountOut);
   const den = api.BigNumber(liquidityOut).minus(amountOut);
   return num.dividedBy(den);
@@ -35,10 +37,13 @@ function getAmountIn(amountOut, liquidityIn, liquidityOut) {
 
 function getAmountOut(amountIn, liquidityIn, liquidityOut) {
   if (!api.assert(api.BigNumber(amountIn).gt(0), 'insufficient output amount')
-    || !api.assert(api.BigNumber(liquidityIn).gt(0) && api.BigNumber(liquidityOut).gt(0), 'insufficient liquidity')) return false;
+    || !api.assert(api.BigNumber(liquidityIn).gt(0)
+      && api.BigNumber(liquidityOut).gt(0), 'insufficient liquidity')) return false;
   const num = api.BigNumber(amountIn).times(liquidityOut);
   const den = api.BigNumber(liquidityIn).plus(amountIn);
-  return num.dividedBy(den);
+  const amountOut = num.dividedBy(den);
+  if (!api.assert(api.BigNumber(amountOut).lte(liquidityOut), 'insufficient liquidity')) return false;
+  return amountOut;
 }
 
 async function validateOracle(pool, newPrice, maxDeviation = 0.01) {
@@ -67,11 +72,14 @@ function validateLiquiditySwap(pool, baseDelta, quoteDelta) {
   return true;
 }
 
-function validateSwap(pool, baseDelta, quoteDelta) {
+function validateSwap(pool, baseDelta, quoteDelta, maxSlippage = 0.01) {
   const k = api.BigNumber(pool.baseQuantity).times(pool.quoteQuantity).toFixed(pool.precision);
   const baseAdjusted = api.BigNumber(pool.baseQuantity).plus(baseDelta);
   const quoteAdjusted = api.BigNumber(pool.quoteQuantity).plus(quoteDelta);
-  // api.debug(`K - ${k}`);
+  const p = api.BigNumber(pool.quoteQuantity).dividedBy(pool.baseQuantity).toFixed(pool.precision);
+  const pAdjusted = api.BigNumber(quoteAdjusted).dividedBy(baseAdjusted).toFixed(pool.precision);
+  const slippage = api.BigNumber(api.BigNumber(pAdjusted - p).abs()).dividedBy(p);
+  if (!api.assert(api.BigNumber(slippage).lte(maxSlippage), 'exceeded max slippage for swap')) return false;
   if (!api.assert(api.BigNumber(api.BigNumber(baseAdjusted).times(quoteAdjusted).toFixed(pool.precision)).eq(k),
     `constant product ${api.BigNumber(baseAdjusted).times(quoteAdjusted)}, expected ${k}`)) return false;
   return true;
@@ -270,18 +278,21 @@ actions.swapTokensForExactTokens = async (payload) => {
     symbolIn = quoteSymbol;
     symbolOut = baseSymbol;
   }
-  if (api.assert(pool, 'no existing pool for tokenPair')) {
-    const tokenInAdjusted = api.BigNumber(getAmountIn(tokenOut, liquidityIn, liquidityOut));
-    const senderBase = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: symbolIn });
-    const senderBaseFunded = api.BigNumber(senderBase.balance).gte(tokenInAdjusted);
-    const tokenPairDelta = tokenSymbol === baseSymbol ? [api.BigNumber(tokenOut).times(-1), tokenInAdjusted] : [tokenInAdjusted, api.BigNumber(tokenOut).times(-1)];
-    if (senderBaseFunded && validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) {
-      await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenInAdjusted.toFixed(pool.precision), to: 'marketpools' });
-      await api.transferTokens(api.sender, symbolOut, tokenOut, 'user');
-      updatePoolStats(pool, tokenPairDelta[0], tokenPairDelta[1]);
-      api.emit('swapTokensForExactTokens', { memo: `Swap ${symbolIn} for ${symbolOut}` });
-    }
-  }
+  if (!api.assert(pool, 'no existing pool for tokenPair')) return;
+
+  const tokenInAdjusted = api.BigNumber(getAmountIn(tokenOut, liquidityIn, liquidityOut));
+  if (!tokenInAdjusted.isFinite()) return;
+
+  const senderBase = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: symbolIn });
+  const senderBaseFunded = api.BigNumber(senderBase.balance).gte(tokenInAdjusted);
+  const tokenPairDelta = tokenSymbol === baseSymbol ? [api.BigNumber(tokenOut).times(-1), tokenInAdjusted] : [tokenInAdjusted, api.BigNumber(tokenOut).times(-1)];
+  if (!api.assert(senderBaseFunded, 'insufficient input balance')
+    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) return;
+
+  await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenInAdjusted.toFixed(pool.precision), to: 'marketpools' });
+  await api.transferTokens(api.sender, symbolOut, tokenOut, 'user');
+  updatePoolStats(pool, tokenPairDelta[0], tokenPairDelta[1]);
+  api.emit('swapTokensForExactTokens', { memo: `Swap ${symbolIn} for ${symbolOut}` });
 };
 
 actions.swapExactTokensForTokens = async (payload) => {
@@ -316,16 +327,20 @@ actions.swapExactTokensForTokens = async (payload) => {
     symbolIn = quoteSymbol;
     symbolOut = baseSymbol;
   }
-  if (api.assert(pool, 'no existing pool for tokenPair')) {
-    const tokenOutAdjusted = api.BigNumber(getAmountOut(tokenIn, liquidityIn, liquidityOut));
-    const senderBase = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: symbolIn });
-    const senderBaseFunded = api.BigNumber(senderBase.balance).gte(tokenIn);
-    const tokenPairDelta = tokenSymbol === baseSymbol ? [tokenIn, api.BigNumber(tokenOutAdjusted).times(-1)] : [api.BigNumber(tokenOutAdjusted).times(-1), tokenIn];
-    if (senderBaseFunded && validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) {
-      await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenIn, to: 'marketpools' });
-      await api.transferTokens(api.sender, symbolOut, tokenOutAdjusted.toFixed(pool.precision), 'user');
-      updatePoolStats(pool, tokenPairDelta[0], tokenPairDelta[1]);
-      api.emit('swapExactTokensForTokens', { memo: `Swap ${symbolIn} for ${symbolOut}` });
-    }
-  }
+
+  if (!api.assert(pool, 'no existing pool for tokenPair')) return;
+
+  const tokenOutAdjusted = api.BigNumber(getAmountOut(tokenIn, liquidityIn, liquidityOut));
+  if (!tokenOutAdjusted.isFinite()) return;
+
+  const senderBase = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: symbolIn });
+  const senderBaseFunded = api.BigNumber(senderBase.balance).gte(tokenIn);
+  const tokenPairDelta = tokenSymbol === baseSymbol ? [tokenIn, api.BigNumber(tokenOutAdjusted).times(-1)] : [api.BigNumber(tokenOutAdjusted).times(-1), tokenIn];
+  if (!api.assert(senderBaseFunded, 'insufficient input balance')
+    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) return;
+
+  await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenIn, to: 'marketpools' });
+  await api.transferTokens(api.sender, symbolOut, tokenOutAdjusted.toFixed(pool.precision), 'user');
+  updatePoolStats(pool, tokenPairDelta[0], tokenPairDelta[1]);
+  api.emit('swapExactTokensForTokens', { memo: `Swap ${symbolIn} for ${symbolOut}` });
 };
--- a/contracts/marketpools.js
+++ b/contracts/marketpools.js
@@ -72,7 +72,7 @@ function validateLiquiditySwap(pool, baseDelta, quoteDelta) {
   return true;
 }
 
-function validateSwap(pool, baseDelta, quoteDelta, maxSlippage = 0.01) {
+function validateSwap(pool, baseDelta, quoteDelta, maxSlippage) {
   const k = api.BigNumber(pool.baseQuantity).times(pool.quoteQuantity).toFixed(pool.precision);
   const baseAdjusted = api.BigNumber(pool.baseQuantity).plus(baseDelta);
   const quoteAdjusted = api.BigNumber(pool.quoteQuantity).plus(quoteDelta);
@@ -143,7 +143,6 @@ actions.create = async (payload) => {
       quoteVolume: 0,
       quotePrice: 0, // Quote per Base (reverse price)
       precision: Math.min(baseToken.precision, quoteToken.precision),
-      active: true,
       creator: api.sender,
     };
     await api.db.insert('pools', newPool);
@@ -267,12 +266,14 @@ actions.swapTokensForExactTokens = async (payload) => {
     tokenPair,
     tokenSymbol,
     tokenOut,
+    maxSlippage,
     isSignedWithActiveKey,
   } = payload;
 
   if (!api.assert(isSignedWithActiveKey === true, 'you must use a transaction signed with your active key')
     || !api.assert(typeof (tokenSymbol) === 'string', 'invalid token')
     || !api.assert(tokenOut && api.BigNumber(tokenOut).gt(0), 'insufficient tokenOut')
+    || !api.assert(maxSlippage && api.BigNumber(maxSlippage).gt(0) && api.BigNumber(maxSlippage).lt(50), 'maxSlippage must be greater than 0 and less than 50')
     || !await validateTokenPair(tokenPair)) {
     return;
   }
@@ -306,7 +307,7 @@ actions.swapTokensForExactTokens = async (payload) => {
   const senderBaseFunded = senderBase && api.BigNumber(senderBase.balance).gte(tokenInAdjusted);
   const tokenPairDelta = tokenSymbol === baseSymbol ? [api.BigNumber(tokenOut).times(-1), tokenInAdjusted] : [tokenInAdjusted, api.BigNumber(tokenOut).times(-1)];
   if (!api.assert(senderBaseFunded, 'insufficient input balance')
-    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) return;
+    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1], api.BigNumber(maxSlippage).dividedBy(100))) return;
 
   const res = await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenInAdjusted.toFixed(pool.precision), to: 'marketpools' });
   if (res.errors === undefined
@@ -322,12 +323,14 @@ actions.swapExactTokensForTokens = async (payload) => {
     tokenPair,
     tokenSymbol,
     tokenIn,
+    maxSlippage,
     isSignedWithActiveKey,
   } = payload;
 
   if (!api.assert(isSignedWithActiveKey === true, 'you must use a transaction signed with your active key')
     || !api.assert(typeof (tokenSymbol) === 'string', 'invalid token')
     || !api.assert(tokenIn && api.BigNumber(tokenIn).gt(0), 'insufficient tokenIn')
+    || !api.assert(maxSlippage && api.BigNumber(maxSlippage).gt(0) && api.BigNumber(maxSlippage).lt(50), 'maxSlippage must be greater than 0 and less than 50')
     || !await validateTokenPair(tokenPair)) {
     return;
   }
@@ -358,7 +361,7 @@ actions.swapExactTokensForTokens = async (payload) => {
   const senderBaseFunded = senderBase && api.BigNumber(senderBase.balance).gte(tokenIn);
   const tokenPairDelta = tokenSymbol === baseSymbol ? [tokenIn, api.BigNumber(tokenOutAdjusted).times(-1)] : [api.BigNumber(tokenOutAdjusted).times(-1), tokenIn];
   if (!api.assert(senderBaseFunded, 'insufficient input balance')
-    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) return;
+    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1], api.BigNumber(maxSlippage).dividedBy(100))) return;
 
   const res = await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenIn, to: 'marketpools' });
   if (res.errors === undefined
--- a/contracts/marketpools.js
+++ b/contracts/marketpools.js
@@ -1,11 +1,13 @@
 /* eslint-disable max-len */
 /* global actions, api */
 
+const TradeType = ['exactInput', 'exactOutput'];
+
 actions.createSSC = async () => {
   const tableExists = await api.db.tableExists('pools');
   if (tableExists === false) {
     await api.db.createTable('pools', ['tokenPair']);
-    await api.db.createTable('liquidityPosition', ['account', 'tokenPair']);
+    await api.db.createTable('liquidityPositions', ['account', 'tokenPair']);
     await api.db.createTable('params');
 
     const params = {};
@@ -96,6 +98,14 @@ async function validateTokenPair(tokenPair) {
   return true;
 }
 
+async function validatePool(tokenPair) {
+  const [baseSymbol, quoteSymbol] = tokenPair.split(':');
+  const pool = await api.db.findOne('pools', { tokenPair });
+  const revPool = await api.db.findOne('pools', { tokenPair: [quoteSymbol, baseSymbol].join(':') });
+  if (!api.assert(pool === null && revPool === null, 'a pool already exists for this tokenPair')) return false;
+  return true;
+}
+
 async function updatePoolStats(pool, baseAdjusted, quoteAdjusted, swap = true) {
   const uPool = pool;
   // precise quantities are needed here for K calculation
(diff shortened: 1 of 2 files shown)


contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+  const {
+    tokenPair,
+    baseQuantity,
+    quoteQuantity,
+    isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(isSignedWithActiveKey === true, 'you must use a transaction signed with your active key')
+    || !api.assert(baseQuantity && api.BigNumber(baseQuantity).gt(0), 'invalid baseQuantity')
+    || !api.assert(quoteQuantity && api.BigNumber(quoteQuantity).gt(0), 'invalid quoteQuantity')
```
bt-cryptomancer, line 171: You should validate that baseQuantity and quoteQuantity are within the precision limits for each token type.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
```
bt-cryptomancer, line 219: You should validate that baseQuantity and quoteQuantity are within the precision limits for each token type.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+      if (api.BigNumber(lp.baseQuantity).eq(0) && api.BigNumber(lp.quoteQuantity).eq(0)) {
+        await api.db.remove('liquidityPosition', lp);
+      } else {
+        await api.db.update('liquidityPosition', lp);
+      }
+
+      await api.transferTokens(api.sender, baseSymbol, baseQuantity, 'user');
+      await api.transferTokens(api.sender, quoteSymbol, quoteQuantity, 'user');
+      updatePoolStats(pool, -baseQuantity, -quoteQuantity, false);
+      api.emit('removeLiquidity', { memo: `Remove ${baseSymbol} and ${quoteSymbol}` });
```
bt-cryptomancer, line 246: Perhaps the emission should also contain the quantities removed?

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+      };
+      await api.db.insert('liquidityPosition', newlp);
+    }
+
+    // deposit requested tokens to contract
+    const baseRes = await api.executeSmartContract('tokens', 'transferToContract', { symbol: baseSymbol, quantity: baseQuantity, to: 'marketpools' });
+    const quoteRes = await api.executeSmartContract('tokens', 'transferToContract', { symbol: quoteSymbol, quantity: quoteQuantity, to: 'marketpools' });
+    if (!api.assert(baseRes.errors === undefined && quoteRes.errors === undefined, 'deposit transfer errors')) return;
+    updatePoolStats(pool, baseQuantity, quoteQuantity, false);
+    api.emit('addLiquidity', { memo: `Add ${baseSymbol} and ${quoteSymbol}` });
```
bt-cryptomancer, line 205: Perhaps the emission should also contain the quantities added?

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+    liquidityOut = pool.quoteQuantity;
+    symbolIn = baseSymbol;
+    symbolOut = quoteSymbol;
+  } else {
+    liquidityIn = pool.quoteQuantity;
+    liquidityOut = pool.baseQuantity;
+    symbolIn = quoteSymbol;
+    symbolOut = baseSymbol;
+  }
+  if (!api.assert(pool, 'no existing pool for tokenPair')) return;
```
bt-cryptomancer, line 283: this assertion should be moved up to right after the findOne statement.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+  const {
+    tokenPair,
+    tokenSymbol,
+    tokenOut,
+    isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(isSignedWithActiveKey === true, 'you must use a transaction signed with your active key')
+    || !api.assert(typeof (tokenSymbol) === 'string', 'invalid token')
+    || !api.assert(tokenOut && api.BigNumber(tokenOut).gt(0), 'insufficient tokenOut')
```
bt-cryptomancer, line 261: You should do validation on the precision of tokenOut.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+    symbolIn = quoteSymbol;
+    symbolOut = baseSymbol;
+  }
+  if (!api.assert(pool, 'no existing pool for tokenPair')) return;
+
+  const tokenInAdjusted = api.BigNumber(getAmountIn(tokenOut, liquidityIn, liquidityOut));
+  if (!tokenInAdjusted.isFinite()) return;
+
+  const senderBase = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: symbolIn });
+  const senderBaseFunded = api.BigNumber(senderBase.balance).gte(tokenInAdjusted);
```
bt-cryptomancer, line 289: Should check that senderBase is not null.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+  const tokenPairDelta = tokenSymbol === baseSymbol ? [api.BigNumber(tokenOut).times(-1), tokenInAdjusted] : [tokenInAdjusted, api.BigNumber(tokenOut).times(-1)];
+  if (!api.assert(senderBaseFunded, 'insufficient input balance')
+    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) return;
+
+  await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenInAdjusted.toFixed(pool.precision), to: 'marketpools' });
+  await api.transferTokens(api.sender, symbolOut, tokenOut, 'user');
```
bt-cryptomancer, line 295: You should put in a sanity check that the transferToContract action succeeded, before transferring tokens to api.sender.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+  const {
+    tokenPair,
+    tokenSymbol,
+    tokenIn,
+    isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(isSignedWithActiveKey === true, 'you must use a transaction signed with your active key')
+    || !api.assert(typeof (tokenSymbol) === 'string', 'invalid token')
+    || !api.assert(tokenIn && api.BigNumber(tokenIn).gt(0), 'insufficient tokenIn')
```
bt-cryptomancer, line 310: You should do validation on the precision of tokenIn.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+    symbolIn = baseSymbol;
+    symbolOut = quoteSymbol;
+  } else {
+    liquidityIn = pool.quoteQuantity;
+    liquidityOut = pool.baseQuantity;
+    symbolIn = quoteSymbol;
+    symbolOut = baseSymbol;
+  }
+
+  if (!api.assert(pool, 'no existing pool for tokenPair')) return;
```
bt-cryptomancer, line 333: this assertion should be moved up to right after the findOne statement.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+
+  const tokenOutAdjusted = api.BigNumber(getAmountOut(tokenIn, liquidityIn, liquidityOut));
+  if (!tokenOutAdjusted.isFinite()) return;
+
+  const senderBase = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: symbolIn });
+  const senderBaseFunded = api.BigNumber(senderBase.balance).gte(tokenIn);
```
bt-cryptomancer, line 339: Should check that senderBase is not null.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+  const tokenPairDelta = tokenSymbol === baseSymbol ? [tokenIn, api.BigNumber(tokenOutAdjusted).times(-1)] : [api.BigNumber(tokenOutAdjusted).times(-1), tokenIn];
+  if (!api.assert(senderBaseFunded, 'insufficient input balance')
+    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1])) return;
+
+  await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: tokenIn, to: 'marketpools' });
+  await api.transferTokens(api.sender, symbolOut, tokenOutAdjusted.toFixed(pool.precision), 'user');
```
bt-cryptomancer, line 345: You should put in a sanity check that the transferToContract action succeeded, before transferring tokens to api.sender.  Also, I question the rounding to pool.precision.  Shouldn't that be to the out token's precision?  If pool.precision > out token precision, this could fail in the tokens contract with a "symbol precision mismatch" error.

bt-cryptomancer (changes requested):
Overall this looks pretty good.  I have some minor questions & quibbles about various validation checks, see comments I've left at specific code lines.

The swapExactTokensForTokens and swapTokensForExactTokens actions use similar code, would it be possible to move that into a helper function that each action calls, in order to avoid code duplication?  You could use a parameter flag to indicate which way the swap should proceed.

For adding and removing liquidity, I think one issue will be how calling applications are supposed to know what are valid amounts of the base & quote tokens that can be added / removed at any given time.  Can you add instructions to your documentation on how a client application can calculate this to perform user input validation before calling the contract actions?  Similarly, I think client apps will want to be able to predict how many tokens will be required as input or received when using the swapExactTokensForTokens and swapTokensForExactTokens actions.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+  if (!api.assert(api.BigNumber(amountOut).lte(liquidityOut), 'insufficient liquidity')) return false;
+  return amountOut;
+}
+
+async function validateOracle(pool, newPrice, maxDeviation = 0.01) {
+  const [baseSymbol, quoteSymbol] = pool.tokenPair.split(':');
+  // eslint-disable-next-line no-template-curly-in-string
+  const baseMetrics = baseSymbol !== "'${CONSTANTS.UTILITY_TOKEN_SYMBOL}$'"
+    ? await api.db.findOneInTable('market', 'metrics', { symbol: baseSymbol })
+    : { lastPrice: 1 };
```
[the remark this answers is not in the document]
donchate: Good catch. I meant this to be CONSTANTS.HIVE_PEGGED_SYMBOL

contracts/marketpools.js
donchate: Should we emit anything if the transferToContract action fails or just return?

contracts/marketpools.js
donchate: My thinking was setting pool precision to the smallest precision of either token in the pair and then using that throughout. It seemed simpler at time of writing to avoid mismatches but maybe there are some cases I didn't consider.

donchate:
I believe frontend apps will have to perform the same math we have in `getAmountOut, getAmountIn` and `validateLiquiditySwap` which shouldn't be too hard with a short explanation and code in documentation I will commit there. Whether there are two swap methods or one, these calculations remain the same so I'm updating the docs while looking into merging them.

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+  const utilityTokenBalance = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: "'${CONSTANTS.UTILITY_TOKEN_SYMBOL}$'" });
+
+  const authorizedCreation = api.BigNumber(poolCreationFee).lte(0) || api.sender === api.owner
+    ? true
+    : utilityTokenBalance && api.BigNumber(utilityTokenBalance.balance).gte(poolCreationFee);
+
+  if (api.assert(authorizedCreation, 'you must have enough tokens to cover the creation fee')
+    && await validateTokenPair(tokenPair)
+    && api.assert(isSignedWithActiveKey === true, 'you must use a transaction signed with your active key')) {
+    const [baseSymbol, quoteSymbol] = tokenPair.split(':');
```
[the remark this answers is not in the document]
donchate: Fixed

contracts/marketpools.js
```diff
@@ -0,0 +1,348 @@
...
+        account: api.sender,
+        tokenPair,
+        baseQuantity,
+        quoteQuantity,
+      };
+      await api.db.insert('liquidityPosition', newlp);
+    }
+
+    // deposit requested tokens to contract
+    const baseRes = await api.executeSmartContract('tokens', 'transferToContract', { symbol: baseSymbol, quantity: baseQuantity, to: 'marketpools' });
```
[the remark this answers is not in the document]
donchate: Indeed, fixed.

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: I was thinking it would be redundant if those amounts will already be emitted in the transferToContract events.

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
```diff
@@ -0,0 +1,370 @@
...
+
+function validateLiquiditySwap(pool, baseDelta, quoteDelta) {
+  const p = api.BigNumber(pool.quoteQuantity).dividedBy(pool.baseQuantity).toFixed(pool.precision);
+  // api.debug(`P - ${p}`);
+  if (!api.assert(api.BigNumber(api.BigNumber(quoteDelta).dividedBy(baseDelta).toFixed(pool.precision)).eq(p),
+    `constant price ${api.BigNumber(quoteDelta).dividedBy(baseDelta)}, expected ${p}`)) return false;
+  return true;
+}
+
+function validateSwap(pool, baseDelta, quoteDelta, maxSlippage = 0.01) {
```
donchate, line 75: Also thinking if we wish to expose the maxSlippage to the frontend to allow the user to risk more than 1% on their swap, if so, upper limit?

donchate:
I am also going to expose maxSlippage and add it to the swap actions payload.

contracts/marketpools.js
bt-cryptomancer: Good point, wasn't thinking about that.  Should be fine as-is then, no change needed.

contracts/marketpools.js
bt-cryptomancer: Should be fine as-is since quantities will be recorded in the transferToContract events.

contracts/marketpools.js
bt-cryptomancer: I think it makes sense to handle it with api.assert so an error is recorded if it fails.  But either way is fine really, as the tokens contract action itself should also generate an error if something goes wrong.  There are examples of doing it both ways in the nft contract, using a helper function to check that the logs generated by the tokens contract action are as expected.

contracts/marketpools.js
bt-cryptomancer: Okay, I forgot that's how the pool precision works.  It should be safe then, as rounding to the pool precision would stay within the precision limits of both tokens.

bt-cryptomancer:
> I believe frontend apps will have to perform the same math we have in `getAmountOut, getAmountIn` and `validateLiquiditySwap` which shouldn't be too hard with a short explanation and code in documentation I will commit there. Whether there are two swap methods or one, these calculations remain the same so I'm updating the docs while looking into merging them.

Okay, sounds good.  Please let me know when you're ready for me to re-review everything.

contracts/marketpools.js
donchate: Both transfers updated to use the corresponding token's precision.

eonwarped:
General comment, can we make sure all rounding directions are consistent for every op? toFixed() should generally truncate or ceiling, not the default rounding mode.

test/marketpools.js
```diff
@@ -0,0 +1,895 @@
+/* eslint-disable */
```
eonwarped, line 1: Can you see if using the fixtures in test/witnesses.js works for you? I've been meaning to refactor all tests to use it. Otherwise don't worry about it much, I'll probably do it all at once.

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+        quoteQuantity,
+      };
+      await api.db.insert('liquidityPositions', newlp);
+    }
+
+    // deposit requested tokens to contract
+    const baseRes = await api.executeSmartContract('tokens', 'transferToContract', { symbol: baseSymbol, quantity: baseQuantity, to: 'marketpools' });
+    const quoteRes = await api.executeSmartContract('tokens', 'transferToContract', { symbol: quoteSymbol, quantity: quoteQuantity, to: 'marketpools' });
+    if (!api.assert(baseRes.errors === undefined && quoteRes.errors === undefined, 'deposit transfer errors')) return;
+    updatePoolStats(pool, baseQuantity, quoteQuantity, false);
```
eonwarped, line 222: Make sure all callers await. Why didn't this fail in test? Make sure there is coverage

donchate: There is coverage, I believe by chance it was completing this call in time. await added to all calls of `updatePoolStats`

bt-cryptomancer (approved):

contracts/marketpools.js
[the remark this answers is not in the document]
bt-cryptomancer: senderBase and senderQuote should have null checks.

bt-cryptomancer (changes requested):
Looks almost perfect, just found one more niggling issue where you forgot to add null checks when doing balance verification with senderBase and senderQuote.  Fix that and it will be good to go!

contracts/marketpools.js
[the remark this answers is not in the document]
donchate: Thanks, added fix and test coverage.

test/marketpools.js
donchate: I would like to refactor for this at a later time.

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+    || !api.assert(api.BigNumber(liquidityIn).gt(0)
+      && api.BigNumber(liquidityOut).gt(0), 'insufficient liquidity')) return false;
+  const num = api.BigNumber(amountIn).times(liquidityOut);
+  const den = api.BigNumber(liquidityIn).plus(amountIn);
+  const amountOut = num.dividedBy(den);
+  if (!api.assert(api.BigNumber(amountOut).lt(liquidityOut), 'insufficient liquidity')) return false;
+  return amountOut;
+}
+
+async function validateOracle(pool, newPrice, maxDeviation = 0.01) {
```
eonwarped, line 51: default for maxDeviation is a float/number type rather than a BigNumber / string. I think it should be "0.01" to reflect exact BigNumber, otherwise floating point representation of 0.01 is imperfect.

test/marketpools.js
eonwarped: sounds good.

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+  const baseMetrics = baseSymbol !== "'${CONSTANTS.HIVE_PEGGED_SYMBOL}$'"
+    ? await api.db.findOneInTable('market', 'metrics', { symbol: baseSymbol })
+    : { lastPrice: 1 };
+  // eslint-disable-next-line no-template-curly-in-string
+  const quoteMetrics = quoteSymbol !== "'${CONSTANTS.HIVE_PEGGED_SYMBOL}$'"
+    ? await api.db.findOneInTable('market', 'metrics', { symbol: quoteSymbol })
+    : { lastPrice: 1 };
+  if (!baseMetrics || !quoteMetrics) return null; // no oracle available
+  const oracle = api.BigNumber(quoteMetrics.lastPrice).dividedBy(baseMetrics.lastPrice);
+  const dev = api.BigNumber(api.BigNumber(newPrice - oracle).abs()).dividedBy(oracle);
```
eonwarped, line 63: newPrice - oracle should be changed to BigNumber subtraction

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+  return true;
+}
+
+function validateSwap(pool, baseDelta, quoteDelta, maxSlippage) {
+  const k = api.BigNumber(pool.baseQuantity).times(pool.quoteQuantity).toFixed(pool.precision);
+  const baseAdjusted = api.BigNumber(pool.baseQuantity).plus(baseDelta);
+  const quoteAdjusted = api.BigNumber(pool.quoteQuantity).plus(quoteDelta);
+  const p = api.BigNumber(pool.quoteQuantity).dividedBy(pool.baseQuantity).toFixed(pool.precision);
+  const pAdjusted = api.BigNumber(quoteAdjusted).dividedBy(baseAdjusted).toFixed(pool.precision);
+  const slippage = api.BigNumber(api.BigNumber(pAdjusted - p).abs()).dividedBy(p);
```
eonwarped, line 83: BigNumber subtraction

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+    const quoteToken = await api.db.findOneInTable('tokens', 'tokens', { symbol: quoteSymbol });
+    const newPool = {
+      tokenPair,
+      baseQuantity: 0,
+      baseVolume: 0,
+      basePrice: 0, // Base per Quote (usual way of reading a pair)
+      quoteQuantity: 0,
+      quoteVolume: 0,
+      quotePrice: 0, // Quote per Base (reverse price)
+      precision: Math.min(baseToken.precision, quoteToken.precision),
```
eonwarped, line 155: Why min here vs max? Or perhaps default should be the highest available? I'm not sure, and will defer to you. But since this appears to be used in intermediate computations throughout the code, larger is better.

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+      };
+      await api.db.insert('liquidityPositions', newlp);
+    }
+
+    // deposit requested tokens to contract
+    const baseRes = await api.executeSmartContract('tokens', 'transferToContract', { symbol: baseSymbol, quantity: baseQuantity, to: 'marketpools' });
+    const quoteRes = await api.executeSmartContract('tokens', 'transferToContract', { symbol: quoteSymbol, quantity: quoteQuantity, to: 'marketpools' });
+    if (!api.assert(baseRes.errors === undefined && quoteRes.errors === undefined, 'deposit transfer errors')) return;
+    await updatePoolStats(pool, baseQuantity, quoteQuantity, false);
+    api.emit('addLiquidity', { memo: `Add ${baseSymbol} and ${quoteSymbol}` });
```
eonwarped, line 223: this deviates from other such events. Can we change this to just  

{ baseSymbol, quoteSymbol }

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+      || !validateLiquiditySwap(pool, baseQuantity, quoteQuantity)) return;
+
+    const lp = await api.db.findOne('liquidityPositions', { account: api.sender, tokenPair });
+    if (api.assert(lp, 'no existing liquidity position')
+      && api.assert(api.BigNumber(lp.baseQuantity).minus(baseQuantity).gte(0)
+      && api.BigNumber(lp.quoteQuantity).minus(quoteQuantity).gte(0), 'not enough liquidity in position')) {
+      lp.baseQuantity = api.BigNumber(lp.baseQuantity).minus(baseQuantity);
+      lp.quoteQuantity = api.BigNumber(lp.quoteQuantity).minus(quoteQuantity);
+      if (api.BigNumber(lp.baseQuantity).eq(0) && api.BigNumber(lp.quoteQuantity).eq(0)) {
+        await api.db.remove('liquidityPositions', lp);
```
eonwarped, line 261: small question-- why not just always update? I get saving space, but seems negligible gain?

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+      } else {
+        await api.db.update('liquidityPositions', lp);
+      }
+
+      await api.transferTokens(api.sender, baseSymbol, baseQuantity, 'user');
+      await api.transferTokens(api.sender, quoteSymbol, quoteQuantity, 'user');
+      await updatePoolStats(pool, -baseQuantity, -quoteQuantity, false);
+      api.emit('removeLiquidity', { memo: `Remove ${baseSymbol} and ${quoteSymbol}` });
```
eonwarped, line 269: same comment about memo vs event with object (so something parsing this doesn't need to parse out the memo if it cares about it)

contracts/marketpools.js
donchate: That should work. The reason I went with a readable string was figured this would go in a column on the frontends.

contracts/marketpools.js
donchate: Original idea was use min for all intermediate computations and when sending out at the end use that for both tokens. But since that we are using individual token precision now on the final send so I believe this could be max now. In the process of reviewing the parts with rounding.

contracts/marketpools.js
donchate: Didn't see a purpose for keeping them once they are removed/destroyed.

contracts/marketpools.js
bt-cryptomancer: I like eonwarped's suggestion here, front-ends can interpret and pretty-print it as they wish.

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Fixed

contracts/marketpools.js
donchate: Made it a BigNumber

contracts/marketpools.js
donchate: Changed

donchate:
> General comment, can we make sure all rounding directions are consistent for every op? toFixed() should generally truncate or ceiling, not the default rounding mode.

Is there a way to set this for the entire contract?

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+    || !api.assert(api.BigNumber(liquidityIn).gt(0)
+      && api.BigNumber(liquidityOut).gt(0), 'insufficient liquidity')) return false;
+  const num = api.BigNumber(amountIn).times(liquidityOut);
+  const den = api.BigNumber(liquidityIn).plus(amountIn);
+  const amountOut = num.dividedBy(den);
+  if (!api.assert(api.BigNumber(amountOut).lt(liquidityOut), 'insufficient liquidity')) return false;
+  return amountOut;
+}
+
+async function validateOracle(pool, newPrice, maxDeviation = api.BigNumber(0.01)) {
```
bt-cryptomancer, line 51: To be absolutely safe, I think this should be written using the string form of the constructor, as maxDeviation = api.BigNumber('0.01')

contracts/marketpools.js
```diff
@@ -0,0 +1,346 @@
...
+    senderBaseFunded = senderBase && api.BigNumber(senderBase.balance).gte(tokenAmountAdjusted);
+    tokenPairDelta = tokenSymbol === baseSymbol ? [api.BigNumber(tokenAmount).times(-1), tokenAmountAdjusted] : [tokenAmountAdjusted, api.BigNumber(tokenAmount).times(-1)];
+    tokenQuantity = { in: tokenAmountAdjusted, out: tokenAmount };
+    if (!api.assert(api.BigNumber(tokenQuantity.out).dp() <= tokenOut.precision, 'symbolOut precision mismatch')) return;
+  }
+
+  if (!api.assert(senderBaseFunded, 'insufficient input balance')
+    || !validateSwap(pool, tokenPairDelta[0], tokenPairDelta[1], api.BigNumber(maxSlippage).dividedBy(100))) return;
+
+  const res = await api.executeSmartContract('tokens', 'transferToContract', { symbol: symbolIn, quantity: api.BigNumber(tokenQuantity.in).toFixed(tokenIn.precision), to: 'marketpools' });
```
bt-cryptomancer, line 339: The senderBaseFunded check is done on the unrounded amount.  The final amount send by the the transferToContract action may be slightly different due to the rounding by toFixed.  You should modify the senderBaseFunded check to be done on the final amount to be sent (with the toFixed rounding).

contracts/marketpools.js
donchate: Set as ```api.BigNumber('0.01')```

contracts/marketpools.js
donchate: I don't think this would be an issue now that we've put ROUND_DOWN in most places but I've added it anyway to be safe.

contracts/marketpools.js
```diff
@@ -0,0 +1,350 @@
...
+
+function validateSwap(pool, baseDelta, quoteDelta, maxSlippage) {
+  const k = api.BigNumber(pool.baseQuantity).times(pool.quoteQuantity).toFixed(pool.precision, api.BigNumber.ROUND_HALF_UP);
+  const baseAdjusted = api.BigNumber(pool.baseQuantity).plus(baseDelta);
+  const quoteAdjusted = api.BigNumber(pool.quoteQuantity).plus(quoteDelta);
+  const p = api.BigNumber(pool.quoteQuantity).dividedBy(pool.baseQuantity).toFixed(pool.precision, api.BigNumber.ROUND_DOWN);
+  const pAdjusted = api.BigNumber(quoteAdjusted).dividedBy(baseAdjusted).toFixed(pool.precision, api.BigNumber.ROUND_DOWN);
+  const slippage = api.BigNumber(pAdjusted).minus(p).abs().dividedBy(p);
+  if (!api.assert(api.BigNumber(slippage).lte(maxSlippage), 'exceeded max slippage for swap')) return false;
+  if (!api.assert(api.BigNumber(api.BigNumber(baseAdjusted).times(quoteAdjusted).toFixed(pool.precision, api.BigNumber.ROUND_HALF_UP)).eq(k),
```
eonwarped, line 85: To make this bullet-proof, I would give it lee-way, e.g.  abs( k.minus( computed ) ) <= 10^{-precision}

eonwarped: oh. Rounding actually is probably equivalent, come to think of it.

contracts/marketpools.js
eonwarped: Oh, this is impossible to be used again, I see now.

contracts/marketpools.js
```diff
@@ -0,0 +1,350 @@
...
+      lp.quoteQuantity = api.BigNumber(lp.quoteQuantity).minus(quoteQuantity);
+      if (api.BigNumber(lp.baseQuantity).eq(0) && api.BigNumber(lp.quoteQuantity).eq(0)) {
+        await api.db.remove('liquidityPositions', lp);
+      } else {
+        await api.db.update('liquidityPositions', lp);
+      }
+
+      await api.transferTokens(api.sender, baseSymbol, baseQuantity, 'user');
+      await api.transferTokens(api.sender, quoteSymbol, quoteQuantity, 'user');
+      await updatePoolStats(pool, -baseQuantity, -quoteQuantity, false);
```
eonwarped, line 272: Change unary minus to BigNumber . negated()

contracts/marketpools.js
```diff
@@ -0,0 +1,350 @@
...
+  if (!baseMetrics || !quoteMetrics) return null; // no oracle available
+  const oracle = api.BigNumber(baseMetrics.lastPrice).dividedBy(quoteMetrics.lastPrice);
+  const dev = api.BigNumber(newPrice).minus(oracle).abs().dividedBy(oracle);
+  // api.debug(`${oracle} -> ${dev} / ${maxDeviation}`);
+  if (!api.assert(api.BigNumber(dev).lte(maxDeviation), 'exceeded max deviation from order book')) return false;
+  return true;
+}
+
+function validateLiquiditySwap(pool, baseDelta, quoteDelta) {
+  const p = api.BigNumber(pool.quoteQuantity).dividedBy(pool.baseQuantity).toFixed(pool.precision, api.BigNumber.ROUND_DOWN);
```
eonwarped, line 70: for price comparison, ROUND_HALF_UP  for these two. (P and P')

contracts/marketpools.js
```diff
@@ -0,0 +1,350 @@
...
+  if (!api.assert(api.BigNumber(api.BigNumber(quoteDelta).dividedBy(baseDelta).toFixed(pool.precision, api.BigNumber.ROUND_DOWN)).eq(p),
+    `constant price ${api.BigNumber(quoteDelta).dividedBy(baseDelta)}, expected ${p}`)) return false;
+  return true;
+}
+
+function validateSwap(pool, baseDelta, quoteDelta, maxSlippage) {
+  const k = api.BigNumber(pool.baseQuantity).times(pool.quoteQuantity).toFixed(pool.precision, api.BigNumber.ROUND_HALF_UP);
+  const baseAdjusted = api.BigNumber(pool.baseQuantity).plus(baseDelta);
+  const quoteAdjusted = api.BigNumber(pool.quoteQuantity).plus(quoteDelta);
+  const p = api.BigNumber(pool.quoteQuantity).dividedBy(pool.baseQuantity).toFixed(pool.precision, api.BigNumber.ROUND_DOWN);
```
eonwarped, line 81: ROUND_HALF_UP for p and p'  (I think a general principle we can use is that for validation usage where we are checking amounts, rounding vs truncation should be used.)

contracts/marketpools.js
```diff
@@ -0,0 +1,350 @@
...
+async function updatePoolStats(pool, baseAdjusted, quoteAdjusted, swap = true) {
+  const uPool = pool;
+  // precise quantities are needed here for K calculation
+  // remainder are statistical and can be rounded (updated for swaps only)
+  uPool.baseQuantity = api.BigNumber(pool.baseQuantity).plus(baseAdjusted);
+  uPool.quoteQuantity = api.BigNumber(pool.quoteQuantity).plus(quoteAdjusted);
+
+  // if all LP is removed, don't update the last price
+  if (uPool.baseQuantity.gt(0) && uPool.quoteQuantity.gt(0)) {
+    uPool.basePrice = api.BigNumber(uPool.quoteQuantity).dividedBy(uPool.baseQuantity).toFixed(pool.precision, api.BigNumber.ROUND_DOWN);
```
eonwarped, line 118: These are for stats / display only, so rounding or not probably doesn't matter? Or these might be used for oracle purposes right? So maybe rounding would be consistent?

donchate: These two prices are not currently used for anything in the contract (display only)

contracts/marketpools.js
donchate: Agreed, changed.

contracts/marketpools.js
donchate: Agreed, changed.

contracts/marketpools.js
donchate: Changed

eonwarped:
Looks good to me.

bt-cryptomancer (approved):
Everything looks good after latest batch of changes.

PR #48, opened 2021-02-22
diesel pools launch revisions

donchate:
- restricts `sharesOut` and `maxSlippage` precision to <= 3
- fix error messages to show actual amount being compared in `validateSwap`
- change to string asserts on user input
- **change `addLiquidity` validation to calculate optimal values within new tolerance parameter**

Should not modify schema or any existing data in the DB.

### DIFF
--- a/contracts/marketpools.js
+++ b/contracts/marketpools.js
@@ -27,6 +27,13 @@ actions.updateParams = async (payload) => {
   await api.db.update('params', params);
 };
 
+function getQuote(amount, liquidityIn, liquidityOut) {
+  if (!api.assert(api.BigNumber(amount).gt(0), 'insufficient amount')
+    || !api.assert(api.BigNumber(liquidityIn).gt(0)
+      && api.BigNumber(liquidityOut).gt(0), 'insufficient liquidity')) return false;
+  return api.BigNumber(amount).times(liquidityOut).dividedBy(liquidityIn);
+}
+
 function getAmountIn(amountOut, liquidityIn, liquidityOut) {
   if (!api.assert(api.BigNumber(amountOut).gt(0), 'insufficient output amount')
     || !api.assert(api.BigNumber(liquidityIn).gt(0)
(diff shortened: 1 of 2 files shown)


contracts/marketpools.js
```diff
@@ -181,14 +180,22 @@ actions.addLiquidity = async (payload) => {
...
-    || !api.assert(baseQuantity && api.BigNumber(baseQuantity).gt(0), 'invalid baseQuantity')
-    || !api.assert(quoteQuantity && api.BigNumber(quoteQuantity).gt(0), 'invalid quoteQuantity')
+    || !api.assert(typeof baseQuantity === 'string' && api.BigNumber(baseQuantity).gt(0), 'invalid baseQuantity')
+    || !api.assert(typeof quoteQuantity === 'string' && api.BigNumber(quoteQuantity).gt(0), 'invalid quoteQuantity')
     || !await validateTokenPair(tokenPair)) return;
 
+  let addSlippage = api.BigNumber('0.01');
+  if (maxSlippage) {
+    if (!api.assert(typeof maxSlippage === 'string'
+      && api.BigNumber(maxSlippage).gt(0) && api.BigNumber(maxSlippage).lt(50), 'maxSlippage must be greater than 0 and less than 50')) return;
```
eonwarped, line 195: Do we want to allow 50?

CodeBull: Without high slippage pools with low liquidity can't be used. So, may be allowing it in the contract is fine, users will make informed decision to change it to whatever % they like.

donchate: All that will happen is they'll potentially deposit 50% less than they intended into the pool.

bt-cryptomancer (changes requested):
Can you add validation for a maximum allowed precision on maxSlippage in both the swapTokens and addLiquidity actions?  I don't think we should ever allow arbitrary precision user input, it should be limited to some reasonable number of digits.

Also, I'm not sure I agree with restricting sharesOut to be an integer only.  I can imagine some cases where there's a pool with very large liquidity, when an integer sharesOut might not offer a fine grained enough demarcation.  Perhaps something like 3 or 4 decimal digits of precision would be better.

donchate:
Applied precision <= 3 for maxSlippage, sharesOut

contracts/marketpools.js
```diff
@@ -199,21 +206,42 @@ actions.addLiquidity = async (payload) => {
...
+        if (!api.assert(api.BigNumber(quoteOptimal).gte(quoteMin), 'exceeded max slippage for adding liquidity')) return;
+        amountAdjusted = [baseQuantity, quoteOptimal];
+      } else {
+        const baseOptimal = getQuote(quoteQuantity, pool.quoteQuantity, pool.baseQuantity).toFixed(baseToken.precision, api.BigNumber.ROUND_HALF_UP);
+        if (api.BigNumber(baseOptimal).lte(baseQuantity)) {
+          if (!api.assert(api.BigNumber(baseOptimal).gte(baseMin), 'exceeded max slippage for adding liquidity')) return;
+          amountAdjusted = [baseOptimal, quoteQuantity];
+        }
+      }
+      const pPrecision = Math.min(baseToken.precision, quoteToken.precision);
```
eonwarped, line 225: Want to re-raise this, but in terms of precision, should it not be by the token unit of this price (either quote or base, not sure which)

donchate:
Pending changes to resolve newly discovered test case.

donchate:
The new optimal values on addLiquidity are calculated using verified math to preserve price. The outputs are also verified in test coverage. The additional validation was failing (incorrectly) due to two stages of rounding and it appears it can be removed entirely.

bt-cryptomancer (approved):
Changes look good.

PR #52, opened 2021-03-11
Tokens safety checks

bt-cryptomancer:
This PR makes the following changes to the tokens contract:

* move the account blacklist to params rather than keeping it hard-coded in the contract
* add list of Hive Engine owned accounts to params
* add a new validation to the create action to prevent accounts other than Hive Engine accounts from being able to create a token with SWAP in the symbol
* eliminate token creation fees for Hive Engine accounts
* clean up unused params settings & code left over from earlier fixes

### DIFF
--- a/contracts/tokens.js
+++ b/contracts/tokens.js
@@ -206,14 +206,8 @@ actions.updateParams = async (payload) => {
   if (enableStakingFee && typeof enableStakingFee === 'string' && !api.BigNumber(enableStakingFee).isNaN() && api.BigNumber(enableStakingFee).gte(0)) {
     params.enableStakingFee = enableStakingFee;
   }
-
-  if (blacklist && typeof blacklist === 'string') {
-    try {
-      params.blacklist = JSON.parse(blacklist);
-    } catch (e) {
-      // don't update if error parsing the blacklist
-      return;
-    }
+  if (blacklist && typeof blacklist === 'object') {
+    params.blacklist = blacklist;
   }
 
   await api.db.update('params', params);
--- a/contracts/tokens.js
+++ b/contracts/tokens.js
@@ -21,6 +21,15 @@ const ACCOUNT_BLACKLIST = {
   'user.dunamu': 1,
 };
 
+// these accounts are allowed to create SWAP.xxx tokens
+const HE_ACCOUNTS = {
+  'hive-engine': 1,
+  'swap-eth': 1,
+  'btc-swap': 1,
+  'graphene-swap': 1,
+  'honey-swap': 1,
+};
+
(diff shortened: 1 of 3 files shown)


contracts/tokens.js
```diff
@@ -88,6 +90,9 @@ actions.createSSC = async () => {
       await cancelBadUnstakes();
```
eonwarped, line 90: Let's clean this code up now that the backfill for unstake is done while we are here

test/tokens.js
```diff
@@ -257,6 +258,19 @@ describe('Tokens smart contract', function () {
       const block1 = res;
       const transactionsBlock1 = block1.transactions;
 
+      console.log(JSON.parse(transactionsBlock1[2].logs).errors[0]);
```
eonwarped, line 261: Clear debugging?

test/tokens.js
```diff
@@ -268,6 +282,118 @@ describe('Tokens smart contract', function () {
...
+  it('updates contract params', (done) => {
+    new Promise(async (resolve) => {
+
+      await loadPlugin(blockchain);
+      database1 = new Database();
+      await database1.init(conf.databaseURL, conf.databaseName);
+
+      let transactions = [];
+      transactions.push(new Transaction(12345678901, 'TXID1233', CONSTANTS.HIVE_ENGINE_ACCOUNT, 'contract', 'update', JSON.stringify(contractPayload)));
+      transactions.push(new Transaction(30896501, 'TXID1236', CONSTANTS.HIVE_ENGINE_ACCOUNT, 'tokens', 'updateParams', '{ "tokenCreationFee": "123" }'));
```
eonwarped, line 305: Check that the params blacklist is populated on deploy before the update params?

eonwarped (commented):
Looks good, just a few things

contracts/tokens.js
bt-cryptomancer: done, removed function and unused params fields

test/tokens.js
bt-cryptomancer: I leave console logging in test cases on purpose, I like seeing what is being output while tests are running.

test/tokens.js
bt-cryptomancer: done

bt-cryptomancer:
I've added a few more things, including:

* add list of Hive Engine owned accounts to params
* only Hive Engine owned accounts can create a token with SWAP in the symbol
* Hive Engine owned accounts don't have to pay token creation fees
* clean up unused params settings & code left over from earlier fixes

eonwarped (approved):
Looks good to me. Not sure we need to unset but it doesn't hurt.

PR #53, opened 2021-03-20
NFT Auction Smart Contract

ali-h:
Allows users to auction NFTs for any Price symbol.

- Fee: 1 BEE
- Max Allowed NFTs at one auction: 50

An Auction can run up to 30 days, and attempts to settle after 24 hours of lead bid. To take the lead a bid must be 5% or more from the previous bid and it resets the 24 hour timer. Bids can be cancelled anytime except 5 minutes before settlement. At any given time owner can cancel the auction or settle it with specific or the current lead bid. Bids can be updated with larger quantity at any time. owner may set a minimum bid and a final price which settles the auction immediately if hit. After settlement, the winner gets NFTs and seller gets his bid. All other bids are returned.

### DIFF
--- a/contracts/nftauction.js
+++ b/contracts/nftauction.js
@@ -30,7 +30,7 @@ actions.createSSC = async () => {
     params.maxExpiryTimeMillis = 2592000000; // 30 days
 
     // max auctions to settle per block
-    params.auctionsPerBlock = 2;
+    params.auctionsPerBlock = 1;
     await api.db.insert('params', params);
   }
 };
--- a/contracts/nftauction.js
+++ b/contracts/nftauction.js
@@ -186,33 +186,25 @@ const settleAuction = async (auction, index = null) => {
 
     const token = await api.db.findOneInTable('tokens', 'tokens', { symbol: priceSymbol });
     const price = api.BigNumber(leadBid.bid);
-    const feePercent = auction.fee / 10000;
-    let fee = price.multipliedBy(feePercent).decimalPlaces(token.precision);
-    if (fee.gt(price)) {
-      fee = price; // unlikely but need to be sure
-    }
+    const feePercent = auction.feePercent / 10000;
+    let fee = price.multipliedBy(feePercent)
+      .decimalPlaces(token.precision, api.BigNumber.ROUND_DOWN);
 
-    let payment = price.minus(fee).decimalPlaces(token.precision);
-    if (payment.lt(0)) {
-      payment = api.BigNumber(0); // unlikely but need to be sure
-    }
+    const payment = price.minus(fee)
+      .decimalPlaces(token.precision, api.BigNumber.ROUND_DOWN);
 
     // check if we have to split the fee into market & agent fees
     const marketParams = await api.db.findOne('marketParams', { symbol });
     let agentFee = api.BigNumber(0);
     if (marketParams && marketParams.officialMarket
-      && marketParams.agentCut !== undefined
-      && marketParams.agentCut > 0 && fee.gt(0)) {
-      const agentFeePercent = marketParams.agentCut / 10000;
-      agentFee = fee.multipliedBy(agentFeePercent).decimalPlaces(token.precision);
-      if (agentFee.gt(fee)) {
-        agentFee = api.BigNumber(fee); // unlikely but need to be sure
-      }
+      && marketParams.agentFeePercent !== undefined
+      && marketParams.agentFeePercent > 0 && fee.gt(0)) {
+      const agentFeePercent = marketParams.agentFeePercent / 10000;
+      agentFee = fee.multipliedBy(agentFeePercent)
+        .decimalPlaces(token.precision, api.BigNumber.ROUND_DOWN);
 
-      fee = fee.minus(agentFee).decimalPlaces(token.precision);
-      if (fee.lt(0)) {
-        fee = api.BigNumber(0); // unlikely but need to be sure
-      }
+      fee = fee.minus(agentFee)
+        .decimalPlaces(token.precision, api.BigNumber.ROUND_DOWN);
     }
 
     // send fee to market account
@@ -221,7 +213,7 @@ const settleAuction = async (auction, index = null) => {
     let isMarketFeePaid = false;
     if (fee.gt(0)) {
       isMarketFeePaid = true;
-      fee = fee.toFixed(token.precision);
+      fee = fee.toFixed(token.precision, api.BigNumber.ROUND_DOWN);
       await api.transferTokens(officialMarketAccount, priceSymbol, fee, 'user');
     }
 
@@ -229,7 +221,7 @@ const settleAuction = async (auction, index = null) => {
     let isAgentFeePaid = false;
     if (agentFee.gt(0)) {
       isAgentFeePaid = true;
-      agentFee = agentFee.toFixed(token.precision);
+      agentFee = agentFee.toFixed(token.precision, api.BigNumber.ROUND_DOWN);
       await api.transferTokens(leadBid.marketAccount, priceSymbol, agentFee, 'user');
     }
 
@@ -288,7 +280,7 @@ actions.create = async (payload) => {
     minBid,
     finalPrice,
     priceSymbol,
-    fee,
+    feePercent,
     expiry,
     isSignedWithActiveKey,
   } = payload;
@@ -303,12 +295,12 @@ actions.create = async (payload) => {
     && minBid && typeof minBid === 'string' && api.BigNumber(minBid).isFinite()
     && finalPrice && typeof finalPrice === 'string' && api.BigNumber(finalPrice).isFinite()
     && priceSymbol && typeof priceSymbol === 'string'
-    && typeof fee === 'number' && fee >= 0 && fee <= 10000 && Number.isInteger(fee)
+    && typeof feePercent === 'number' && feePercent >= 0 && feePercent <= 10000 && Number.isInteger(feePercent)
     && expiry && typeof expiry === 'string', 'invalid params')
     && api.assert(nfts.length <= MAX_NUM_UNITS_OPERABLE, `cannot process more than ${MAX_NUM_UNITS_OPERABLE} NFT instances at once`)) {
     const marketParams = await api.db.findOne('marketParams', { symbol });
-    if (marketParams && marketParams.minFee !== undefined) {
-      if (!api.assert(fee >= marketParams.minFee, `fee must be >= ${marketParams.minFee}`)) {
+    if (marketParams && marketParams.minFeePercent !== undefined) {
+      if (!api.assert(feePercent >= marketParams.minFeePercent, `feePercent must be >= ${marketParams.minFeePercent}`)) {
         return;
       }
     }
@@ -385,7 +377,7 @@ actions.create = async (payload) => {
                 priceSymbol,
                 minBid,
                 finalPrice,
-                fee,
+                feePercent,
                 expiryTimestamp,
                 bids: [],
                 currentLead: null,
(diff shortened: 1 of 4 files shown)


contracts/nftauction.js
```diff
@@ -0,0 +1,576 @@
...
+      fromType: 'contract',
+      to: finalTo,
+      toType: 'user',
+      nfts: [wrappedNfts],
+      isSignedWithActiveKey: true,
+    });
+
+    let count = 0;
+
+    while (count < bids.length) {
```
eonwarped, line 147: prefer for loop here since you are always looping through everything

contracts/nftauction.js
```diff
@@ -0,0 +1,576 @@
...
+    && el.data.to === to
+    && api.BigNumber(el.data.quantity).eq(quantity)
+    && el.data.symbol === symbol) !== undefined) {
+    return true;
+  }
+
+  return false;
+};
+
+const settleAuction = async (auction, id = null, settle = true) => {
```
eonwarped, line 104: settle and cancel are different enough that I think this serves better as splitting into their respective functions and extracting common pieces as helper methods. For example, it looks like you return nfts to the seller twice on cancel.

contracts/nftauction.js
```diff
@@ -0,0 +1,576 @@
...
+    isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(symbol && typeof symbol === 'string', 'invalid symbol')) return;
+
+  const nft = await api.db.findOneInTable('nft', 'nfts', { symbol });
+
+  if (api.assert(isSignedWithActiveKey === true, 'you must use a custom_json signed with your active key')
+    && api.assert(nfts && typeof nfts === 'object' && Array.isArray(nfts)
+      && minBid && typeof minBid === 'string' && !api.BigNumber(minBid).isNaN()
```
eonwarped, line 216: probably want to check isFinite()  (can replace isNaN)

contracts/nftauction.js
```diff
@@ -0,0 +1,576 @@
...
+    priceSymbol,
+    expiry,
+    isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(symbol && typeof symbol === 'string', 'invalid symbol')) return;
+
+  const nft = await api.db.findOneInTable('nft', 'nfts', { symbol });
+
+  if (api.assert(isSignedWithActiveKey === true, 'you must use a custom_json signed with your active key')
```
eonwarped, line 214: I think cleaner to just keep using !api.assert(...) return;  as above  so you don't have to nest the following code

contracts/nftauction.js
```diff
@@ -0,0 +1,576 @@
...
+    params.cancelLockTime = 300000; // 5 mins
+
+    // time after the last lead bid it takes to settle the auction
+    params.expiryTime = 86400000; // 24 hours
+
+    // max time an auction can run
+    params.maxExpiryTime = 2592000000; // 30 days
+
+    // max auctions to settle per block
+    params.auctionsPerBlock = 2;
```
eonwarped, line 32: Make sure this parameter is tested

eonwarped (commented):
More comments to come, partial review. (need to look at bids and cancelling bids)

contracts/nftauction.js
```diff
@@ -0,0 +1,597 @@
...
+
+  const blockDate = new Date(`${api.hiveBlockTimestamp}.000Z`);
+  const timestamp = blockDate.getTime();
+
+  const lastValidLead = timestamp - params.expiryTimeMillis;
+
+  const auctionsToSettle = await api.db.find('auctions',
+    {
+      $or: [
+        { lastLeadUpdate: { $lte: lastValidLead } },
```
eonwarped, line 585: for these, you should add indexes for 'lastLeadUpdate'  and 'expiryTimestamp' even though you do not sort on them

contracts/nftauction.js
```diff
@@ -0,0 +1,597 @@
...
+    expiry,
+    isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(symbol && typeof symbol === 'string', 'invalid symbol')) return;
+  if (!api.assert(isSignedWithActiveKey === true, 'you must use a custom_json signed with your active key')) return;
+
+  const nft = await api.db.findOneInTable('nft', 'nfts', { symbol });
+
+  if (api.assert(nfts && typeof nfts === 'object' && Array.isArray(nfts)
```
eonwarped, line 234: here and below: i think can be cleaner if we use early exit as in line 229:  if (!api.assert(...)) return

contracts/nftauction.js
```diff
@@ -0,0 +1,597 @@
...
+
+  // return NFTs to the seller
+  await sendNfts(seller, wrappedNfts);
+
+  api.emit('cancelAuction', {
+    auctionId,
+  });
+};
+
+const settleAuction = async (auction, id = null) => {
```
eonwarped, line 154: can we name 'id' here as 'index', so it's clear it's not some numeric identifier on the bid

test/nftauction.js
```diff
@@ -0,0 +1,1270 @@
...
+
+      await send(blockchain.PLUGIN_NAME, 'MASTER', { action: blockchain.PLUGIN_ACTIONS.PRODUCE_NEW_BLOCK_SYNC, payload: block });
+
+      let res = await database1.getLatestBlockInfo();
+      let txs = res.transactions;
+
+      assertError(txs[10], 'you must use a custom_json signed with your active key');
+      assertError(txs[11], 'invalid params'); // invalid auctionId
+      assertError(txs[12], 'auction does not exist or has been expired');
+      assertError(txs[13], 'you don not have a bid in this auction');
```
eonwarped, line 502: there's a typo in this error (don not --> do not)

test/nftauction.js
```diff
@@ -0,0 +1,1270 @@
...
+      resolve();
+    })
+      .then(() => {
+        unloadPlugin(blockchain);
+        database1.close();
+        done();
+      });
+  });
+
+  it('cancels a bid', (done) => {
```
eonwarped, line 531: does this make sure to test both cases where:

1)  cancels a bid with index < lead bid
2) cancels a bid with index > lead bid

and makes sure the lead is maintained as expected

test/nftauction.js
```diff
@@ -0,0 +1,1270 @@
...
+      resolve();
+    })
+      .then(() => {
+        unloadPlugin(blockchain);
+        database1.close();
+        done();
+      });
+  });
+
+  it('bids multiple times in an auction', (done) => {
```
eonwarped, line 409: i don't see a test case for if the same bidder updates a bid?

eonwarped (commented):
I think this looks good in general. some comments

test/nftauction.js
ali-h: Added a new block in the test unit to confirm (with both cases) and a function that confirms that the index is as expected

test/nftauction.js
ali-h: Added, negative and positive both

eonwarped (dismissed):

eonwarped (dismissed):

bt-cryptomancer:
During testing, reazuliqbal found the following points of concern.  Please resolve each of these:

- Seller can bid on their own auction
- Cancelled auctions are not removed from the database (also no indication that they are canceled) but NFTs are sent back to the seller
- Cancelled auctions can receive bids
- After cancellation its possible to have multiple auctions for the same NFTs, so very hard to distinguish which is cancelled which are not.
- Its possible to settle a cancelled auction if someone bids on it, the contract sends the NFT to the bidder even tho it does not have it.

Also, Aggroed has indicated a preference to have the auction contract include market fees similar to the NFT market contract.  This is especially important for NFT Showroom, and in particular it should obey the officialMarket, agentCut, and minFee that are configured for the NFT in question.  See documentation here for details: https://github.com/hive-engine/steemsmartcontracts-wiki/blob/master/NFT-Market-Contract.md#configuring-the-market

ali-h:
### Update:
- Sellers now can not bid on their own auction
- Canceled auctions now get deleted also a fix in test file which now confirms it
- Added a fee mechanism similar to `nftmarket` contract
- nft issuers can set `marketParams` that define official market account, agent fee cut and minimum fee
- separate test unit for fee mechanism, market params etc

contracts/nftauction.js
```diff
@@ -180,15 +184,66 @@ const settleAuction = async (auction, index = null) => {
     // return the bids
     await returnBids(bids, priceSymbol);
 
-    // send payment to the seller
-    await api.transferTokens(seller, priceSymbol, leadBid.bid, 'user');
+    const token = await api.db.findOneInTable('tokens', 'tokens', { symbol: priceSymbol });
+    const price = api.BigNumber(leadBid.bid);
+    const feePercent = auction.fee / 10000;
+    let fee = price.multipliedBy(feePercent).decimalPlaces(token.precision);
+    if (fee.gt(price)) {
```
eonwarped, line 191: Don't use this check, instead explicitly set the rounding mode to ROUND_DOWN

eonwarped: This applies to all calls to decimalPlaces(), picking the rounding mode that makes sense for the operation

contracts/nftauction.js
```diff
@@ -180,15 +184,66 @@ const settleAuction = async (auction, index = null) => {
...
+      }
+    }
+
+    // send fee to market account
+    const officialMarketAccount = (marketParams && marketParams.officialMarket)
+      ? marketParams.officialMarket : leadBid.marketAccount;
+    let isMarketFeePaid = false;
+    if (fee.gt(0)) {
+      isMarketFeePaid = true;
+      fee = fee.toFixed(token.precision);
```
eonwarped, line 224: calls to toFixed also should explicitly set rounding mode

contracts/nftauction.js
```diff
@@ -222,6 +288,7 @@ actions.create = async (payload) => {
     minBid,
     finalPrice,
     priceSymbol,
+    fee,
```
eonwarped, line 291: feePercent, variable should indicate some form of units. Documentation will clarify that it is percentage*100 , max 10000

Ditto for market parameter

contracts/nftauction.js
```diff
@@ -599,3 +688,82 @@ actions.updateAuctions = async () => {
...
+
+      if (shouldUpdate) {
+        if (isFirstTimeSet) {
+          await api.db.insert('marketParams', params);
+        } else {
+          await api.db.update('marketParams', params);
+        }
+
+        api.emit('setMarketParams', update);
+        return true;
```
eonwarped, line 764: This emit is unnecessary, it can be inferred from no error in operation

contracts/nftauction.js
```diff
@@ -599,3 +688,82 @@ actions.updateAuctions = async () => {
...
+  const {
+    symbol, officialMarket, agentCut, minFee, isSignedWithActiveKey,
+  } = payload;
+
+  if (!api.assert(symbol && typeof symbol === 'string', 'invalid params')) {
+    return false;
+  }
+
+  // if no parameters supplied, we have nothing to do
+  if (officialMarket === undefined && agentCut === undefined && minFee === undefined) {
```
eonwarped, line 702: There appears to be a lot of extra logic here. 

Recommendation is just to create the params up front if it does not exist,
Then go through the settings and update.

Also, I would assert errors so it's obvious if a provided parameter is malformed.

It looks like you are distinguishing between null and undefined as well, so people can unset a parameter right?

ali-h:
### Update:
- Set the rounding mode to ROUND_DOWN in decimalPlaces and toFixed
- Rename fee parameters in for better understanding (feePercent, minFeePercent, agentFeePercent)
- Remove the extra logic from setMarketParams action

eonwarped (approved):
Looks good to me

PR #55, opened 2021-03-22
marketpools randomized test case bounds fix

donchate:
The randomized test case was sometimes overdrawing the available balance and returning a false negative result. The amounts are scaled back and should perform reliably going forward. I found this test useful during development, but perhaps it doesn't need to be run in the production test suite at all.


bt-cryptomancer (approved):
changes look good

PR #56, opened 2021-03-23
Decentralized (or Distributed) Token Fund (DTF)

donchate:
Docs at https://github.com/donchate/steemsmartcontracts-wiki/blob/marketpools/TokenFund-Contract.md
Docs PR https://github.com/hive-engine/steemsmartcontracts-wiki/pull/10

Hive block number for activation needs to be set after reviews are complete.


contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+  }
+  if (processQueryLimit) {
+    if (!api.assert(typeof processQueryLimit === 'string' && api.BigNumber(processQueryLimit).isInteger() && api.BigNumber(processQueryLimit).gte(1), 'invalid processQueryLimit')) return;
+    params.processQueryLimit = api.BigNumber(processQueryLimit).toNumber();
+  }
+  await api.db.update('params', params);
+};
+
+function generateDtfId(dtf) {
+  return `${dtf.payToken.replace('.', '-')}:${dtf.voteToken.replace('.', '-')}`;
```
bt-cryptomancer, line 63: What is the purpose of this function?  Why do you need to replace '.' with '-' ?

contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+
+async function updateProposalWeight(id, deltaApprovalWeight) {
+  const proposal = await api.db.findOne('proposals', { _id: id });
+  if (proposal) {
+    proposal.approvalWeight = api.BigNumber(proposal.approvalWeight).plus(deltaApprovalWeight).toNumber();
+    await api.db.update('proposals', proposal);
+  }
+}
+
+async function validateTokens(payTokenObj, voteTokenObj) {
```
bt-cryptomancer, line 74: This function doesn't need to be async, as it doesn't await anything.

contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+    && api.assert(typeof maxAmountPerDay === 'string' && api.BigNumber(maxAmountPerDay).gt(0), 'invalid maxAmountPerDay: greater than 0')) {
+    if (proposalFee) {
+      if (!api.assert(typeof proposalFee === 'object'
+        && typeof proposalFee.method === 'string' && FeeMethod.indexOf(proposalFee.method) !== -1
+        && typeof proposalFee.symbol === 'string'
+        && typeof proposalFee.amount === 'string' && api.BigNumber(proposalFee.amount).gt(0), 'invalid proposalFee')) return;
+      const feeTokenObj = await api.db.findOneInTable('tokens', 'tokens', { symbol: proposalFee.symbol });
+      if (!api.assert(feeTokenObj && api.BigNumber(proposalFee.amount).dp() <= feeTokenObj.precision, 'invalid proposalFee token or precision')) return;
+    }
+    const existingDtf = await api.db.findOne('funds', { id: fundId });
```
bt-cryptomancer, line 207: If the generateDtfId function is required, then should fundId be generated through a similar process so external callers won't have to be concerned with that detail?  Same thing applies to other actions that have a fundId parameter.

contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+  const { dtfUpdateFee } = params;
+
+  // eslint-disable-next-line no-template-curly-in-string
+  const utilityTokenBalance = await api.db.findOneInTable('tokens', 'balances', { account: api.sender, symbol: "'${CONSTANTS.UTILITY_TOKEN_SYMBOL}$'" });
+
+  const authorizedCreation = api.BigNumber(dtfUpdateFee).lte(0) || api.sender === api.owner
+    ? true
+    : utilityTokenBalance && api.BigNumber(utilityTokenBalance.balance).gte(dtfUpdateFee);
+
+  if (api.assert(authorizedCreation, 'you must have enough tokens to cover the creation fee')
```
bt-cryptomancer, line 194: creation fee should say **update** fee

contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+    dtf.active = !!active;
+    await api.db.update('funds', dtf);
+    api.emit('setDtfActive', { id: dtf.id, active: dtf.active });
+  }
+};
+
+actions.createProposal = async (payload) => {
+  const {
+    fundId, title, startDate, endDate, amountPerDay,
+    authorperm, payout, isSignedWithActiveKey,
```
bt-cryptomancer, line 253: authorperm feels to me like it's short for "author permissions".
Can you change that to something like authorpermlink, or aboutlink, to clarify the intent?

bt-cryptomancer (changes requested):
Mostly looks good, just some minor revisions needed.  See comments.

contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+  const fundedLog = [];
+  let offset = 0;
+  let proposals;
+  let runningPay = api.BigNumber(dtf.maxAmountPerDay).times(tickPayRatio);
+  while (runningPay.gt(0)) {
+    proposals = await api.db.find('proposals',
+      {
+        fundId: dtf.id,
+        active: true,
+        approvalWeight: { $gt: api.BigNumber(dtf.voteThreshold).toNumber() },
```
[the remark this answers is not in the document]
donchate: Updated.

contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+};
+
+function generateDtfId(dtf) {
+  return `${dtf.payToken.replace('.', '-')}:${dtf.voteToken.replace('.', '-')}`;
+}
+
+async function updateProposalWeight(id, deltaApprovalWeight) {
+  const proposal = await api.db.findOne('proposals', { _id: id });
+  if (proposal) {
+    proposal.approvalWeight = api.BigNumber(proposal.approvalWeight).plus(deltaApprovalWeight).toNumber();
```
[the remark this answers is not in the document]
donchate: Updated.

contracts/tokenfunds.js
donchate: Fixed.

contracts/tokenfunds.js
donchate: Inherited from mining, its not required here. Removed.

contracts/tokenfunds.js
donchate: Fixed.

contracts/tokenfunds.js
donchate: authorperm naming came from hived, updated to authorPermlink to more suit our naming style

contracts/tokenfunds.js
```diff
@@ -0,0 +1,583 @@
...
+      acct.weights.push({ symbol: token.symbol, weight: approvalWeight });
+    }
+
+    const deltaApprovalWeight = api.BigNumber(approvalWeight)
+      .minus(oldApprovalWeight)
+      .dp(token.precision, api.BigNumber.ROUND_HALF_UP);
+
+    if (!api.BigNumber(deltaApprovalWeight).eq(0)) {
+      await api.db.update('accounts', acct);
+      const approvals = await api.db.find('approvals', { from: account });
```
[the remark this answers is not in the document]
donchate: I have added a maxAccountApprovals param to scope this query to the last 1000 added approvals.

donchate:
I spent probably more time than necessary thinking about retention of approvals and proposal weights after they are completed. \

Now, we cap at 1000 latest approvals for users, and after proposals are done the proposal.approvalweight remains static and is no longer updated. If we need to clean up more aggressively, the next step would be to remove records from the approvals table, which I think would be a substantial negative for users and frontends that wish to display a list of user approvals and past proposal support, etc. Also, I believe this would require an updated remove query in the database api to allow multiple record removal via query rather than one at a time.

contracts/tokenfunds.js
```diff
@@ -0,0 +1,601 @@
...
+    if (!api.BigNumber(deltaApprovalWeight).eq(0)) {
+      await api.db.update('accounts', acct);
+      const approvals = await api.db.find('approvals',
+        { from: account },
+        params.maxAccountApprovals,
+        0,
+        [{ index: '_id', descending: true }]);
+      for (let index = 0; index < approvals.length; index += 1) {
+        const approval = approvals[index];
+        await updateProposalWeight(approval.to, deltaApprovalWeight);
```
bt-cryptomancer, line 514: I'm still a bit concerned about performance in this loop.  maxAccountApprovals is 1000, so in the worst case we might call updateProposalWeight 1000 times, and that function has a both a DB find & update, so again in the worst case that's 2000 more queries, though most likely the validation check against the proposal's endDate will cut down the updates significantly.

You could optimize it a bit by copying the endDate into the approvals table when the approval is created.  That way the information will already be available here so you can do the endDate validation without performing another DB find if the proposal has already ended.  Although if you made that change, you'd probably also want to remove the ability to update the endDate through the updateProposal action, since that would then require updating it in all the approvals table entries as well, which could have its own performance implications.  So there are trade-offs.

Also, maybe a smaller number for maxAccountApprovals would be appropriate?  1000 seems like a lot, perhaps 100 or even 50 would be better?

bt-cryptomancer (changes requested):
Still a couple things for you to look at before I approve this, but almost done!

> If we need to clean up more aggressively, the next step would be to remove records from the approvals table, which I think would be a substantial negative for users and frontends that wish to display a list of user approvals and past proposal support, etc. Also, I believe this would require an updated remove query in the database api to allow multiple record removal via query rather than one at a time.

Fair enough, and good point about multiple record removal.  We can leave the records in the table for now and see if DB storage space ever starts becoming enough of an issue to think about methods for cleaning it up.  I think it will be a while before we really need to worry about this.

contracts/tokenfunds.js
donchate: In my experience with the hived/DHF there have been quite a few requests for reducing the end date via an update. Proposal creators gain more flexibility if they are unsure whether they will get funded or not, and after successful funding can adjust amount and end to meet their goals without having to create a new one. It would be good to try and keep it.

An advantage of going with fewer though, such as 30 or 50, should encourage people to approve only what they believe to be higher quality proposals. It is designed to limit approvals on active proposals at any given time, so instead of having updateProposal do a large query to do the update, I am trying to update them as they go along in approveProposal and updateProposalApprovals.

Also noticed here we need to enforce the max when people are adding new approvals, to prevent any stake against an active proposal to not be updated. So I need to add more queries here, but they offer a chance to update approvals on finished proposals and verify user is not going over the max amount. If people come against this error when approving, then it will be time to consider increasing maxAccountApprovals.

So, going with 50, in either case I think it would be a max of 50 + 50 queries. (The endDate validation now doesn't save a query,  but causes the approval to be updated instead, but allows the whole thing to remain consistent).

bt-cryptomancer (approved):
> In my experience with the hived/DHF there have been quite a few requests for reducing the end date via an update. Proposal creators gain more flexibility if they are unsure whether they will get funded or not, and after successful funding can adjust amount and end to meet their goals without having to create a new one. It would be good to try and keep it.

Okay, this is pretty compelling reasoning, I agree it's beneficial to keep the end date updateable.

> An advantage of going with fewer though, such as 30 or 50, should encourage people to approve only what they believe to be higher quality proposals. It is designed to limit approvals on active proposals at any given time, so instead of having updateProposal do a large query to do the update, I am trying to update them as they go along in approveProposal and updateProposalApprovals.

> Also noticed here we need to enforce the max when people are adding new approvals, to prevent any stake against an active proposal to not be updated. So I need to add more queries here, but they offer a chance to update approvals on finished proposals and verify user is not going over the max amount. If people come against this error when approving, then it will be time to consider increasing maxAccountApprovals.

An approval limit of 50 feels good, and I like the approach you're taking here which should ultimately work a bit better.

I'm happy to approve this now, great work!

PR #57, opened 2021-03-24
Index changes for use in new contracts

eonwarped:
Add ability to designate a primary key to take advantage of mongo's _id indexing.
Add ability to designate a compound index when defining a table.
Added example add index for mining contract.

Note: indexes cannot be added within a transaction, so 'addIndex' functionality is added without session, and would be available in the next Transaction. (Even now, indexes created in the creation are not available within the same transaction for a similar reason). 

Primary keys will be automatically referenced in find queries, and will speed up the operation if any primary key prefix is specified in the query.

Compound  indexes will also take effect automatically in find operations, and can be specified in sorts by name. See comments contract for example usage: https://github.com/hive-engine/steemsmartcontracts/pull/54


libs/Database.js
```diff
@@ -382,15 +392,18 @@ class Database {
    * @param {Array} indexes array of string containing the name of the indexes to create
```
bt-cryptomancer, line 392: should add a doc string here for the new params parameter

libs/Database.js
```diff
@@ -462,14 +534,39 @@ class Database {
         const finalTableName = `${contract}_${table}`;
         const tableData = await this.getContractCollection(contract, finalTableName);
 
+        const contractInDb = await this.findContract({ name: contract });
```
bt-cryptomancer, line 537: getContractCollection immediately above also does a findContract.  To avoid doing findContract twice, it would be more efficient to modify getContractCollection to return an object with both the tableData and primaryKey.

libs/Database.js
```diff
@@ -531,7 +639,6 @@ class Database {
             await this.flushCache();
           }
         }
-        const finalTableName = `${contract}_${table}`;
 
         const tableData = await this.getContractCollection(contract, finalTableName);
```
bt-cryptomancer, line 643: same thing here, can avoid doing findContract twice if getContractCollection also returns the primaryKey

bt-cryptomancer (changes requested):
Looks good aside from just a few very minor points, see comments.

libs/Database.js
eonwarped: done

eonwarped (commented):
resolved comments

bt-cryptomancer (approved):
Changes look good, thanks!

PR #58, opened 2021-04-03
diesel pools: add action to update reward pool

donchate:
Quick update to make it possible for a pool's reward lottery settings to be updated.
Fee set to match mining contract update fee.


contracts/marketpools.js
```diff
@@ -13,17 +13,27 @@ actions.createSSC = async () => {
     const params = {};
     params.poolCreationFee = '1000';
     await api.db.insert('params', params);
+  } else {
+    const params = await api.db.findOne('params', {});
+    if (!params.poolUpdateFee) {
+      params.poolUpdateFee = '300';
```
bt-cryptomancer, line 19: The mining pool update fee was recently changed to 100.  Perhaps you should change this to match?

bt-cryptomancer (changes requested):
It looks like fees will get applied twice (once by the marketpools contract and again by the mining contract).  So the mining contract's update action should be changed to not apply its fee if being called from marketpools.

Also I think you should change params.poolUpdateFee to match the mining contract (see comment below in code).

As this includes @eonwarped 's index changes, I will wait to merge until after the index change PR is merged, and we can deploy following the next node release.

contracts/marketpools.js
donchate: Replaced reward pool fee check with those from mining contract params. Maintaining separate fees for a simple wrapper seems unnecessary at this time and I want to avoid more changes to mining contract itself.

bt-cryptomancer: Sounds sensible, I'm fine with this approach, thanks.

bt-cryptomancer (approved):
Changes look good now, thanks.

PR #61, opened 2021-04-04
fix start block detection and skip block detection

eonwarped:
Now it will take into account the additionalHashChangeBlocks, which is the scenario where for some reason or other the node is processing a hive block that changes the DB *without* producing any logs / additional block (e.g. with the automated actions). This can happen right now because of comment actions that get sent to the Blockchain plugin but get ignored due contract not being implemented and then subsequently getting filtered out.

Note that this is actually not so much of a risk because most such actions modifying the DB are idempotent (they will only happen once based on a timestamp condition, so re-running the same action again will not produce additional differences). However, this may not always be the case, so we should be proactive about preventing a state where a node restart would apply double-actions inadvertently.

### DIFF
--- a/plugins/Blockchain.js
+++ b/plugins/Blockchain.js
@@ -42,6 +42,13 @@ function addBlock(block) {
   return database.addBlock(block);
 }
 
+function getRefBlockNumber(block) {
+  if (block.otherHashChangeRefHiveBlocks) {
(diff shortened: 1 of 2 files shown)


bt-cryptomancer (approved):
changes look good

PR #64, opened 2021-05-04
Fix logging of witness_actions

Rishi556:
No response was given previously. This fixes that and has the action give a response.


bt-cryptomancer (approved):
Changes look good.

PR #69, opened 2021-05-22
gate new streamer behavior on forking block

eonwarped:
Allows 1.4.0 to replay from snapshots prior to the inadvertent fork.


bt-cryptomancer (approved):
This looks fine, can include in the next release.

PR #70, opened 2021-05-23
add witness updateParams action

eonwarped:
updateParams will auto reset schedule if the witness number differs.

Also, cleans up old gates that are no longer used.


bt-cryptomancer (approved):
Changes look good.

PR #71, opened 2021-05-23
Clean up post vote data after payout

eonwarped:
Cleans up post and vote data after payout, as it is no longer needed.

Emit event for declined post payout so indexers can pick up the change as well.


bt-cryptomancer (approved):
Changes look good.

PR #74, opened 2021-06-04
Fix mining contract power update. Ensure on pool update we do a full recomputation

eonwarped:


bt-cryptomancer (approved):
Changes look good.

PR #75, opened 2021-06-06
fix comment vote queries

eonwarped:
Testing this on the testnet, should speed up queries during post payout processing.


bt-cryptomancer (approved):
Changes look good.

PR #77, opened 2021-06-22
Rounding outputs for tokens of smaller precision

donchate:
Essentially changing api.BigNumber.ROUND_HALF_UP to api.BigNumber.ROUND_DOWN on the output from the contract. Also minor code cleanup for readability and test case.


bt-cryptomancer (approved):
Changes look good.

PR #81, opened 2021-07-06
Add tick registration mechanism

eonwarped:
Register all ticking contracts into "contracts_config" table. Allows for putting more global contract settings in the future. Add "registerTick" method for registering additional ticking actions.

### DIFF
--- a/contracts/testing/dummyTicks.js
+++ b/contracts/testing/dummyTicks.js
@@ -1,0 +1,27 @@
+actions.createSSC = function (payload) {
+  // Initialize the smart contract via the create action
+}
+actions.checkPendingUnstakes = () => {
+    api.emit('checkPendingUnstakes', {});
+}
+actions.checkPendingUndelegations = () => {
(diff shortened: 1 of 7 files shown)


bt-cryptomancer (changes requested):
This looks great, should work well!  I like how the registerTick action sets the starting block to be the block after it's registered in; that will make things faster for the next release as we can simply wait until all witnesses have upgraded and then register the new tick for donchate's contract.

Some minor points of polish to look at:

1. contractsConfig can also be used to simplify the big chain of if-else-if for "unknown errors" starting around line 164 of libs/Block.js.

2. The registerTick action returns empty logs if it succeeds. I think it should generate some logging output with the contract, action, and startRefBlock.

3. Please fix the lint errors.

eonwarped:
Requested changes done. PTAL

bt-cryptomancer (approved):
Thanks, revisions look good.

PR #82, opened 2021-07-06
Add chain id to getStatus

Rishi556:
Adds chain id from config to getStatus


bt-cryptomancer (changes requested):
This looks fine, but please remove package-lock.json from the PR.  I realize we forgot to bump up the version number in that file, will do it separately when we prepare the next release.

bt-cryptomancer (approved):
changes look good

PR #84, opened 2021-07-27
NFT mining equip fields and extra mining power fields

eonwarped:
Allow specifying equip field or mining power field to allow NFT creator to control what is considered as equipped and allow for extra mining power to be added per nft instance.

These can configured independently, and cannot be changed once a reward pool is created.

We add a new contract call from nft contract to mining contract whenever setProperties is called, which allows for the corresponding mining power to be updated.

### DIFF
--- a/contracts/nft.js
+++ b/contracts/nft.js
@@ -960,7 +960,9 @@ actions.setProperties = async (payload) => {
               shouldUpdate = true;
             }
             if (shouldUpdate) {
-              await api.executeSmartContract('mining', 'handleNftSetProperty', { symbol, nft: nftInstance, propertyName: name, oldValue: oldProperty });
+              await api.executeSmartContract('mining', 'handleNftSetProperty', {
+                symbol, nft: nftInstance, propertyName: name, oldValue: oldProperty,
+              });
             }
           }
           if (shouldUpdate) {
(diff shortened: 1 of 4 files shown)


contracts/nft.js
```diff
@@ -947,16 +947,23 @@ actions.setProperties = async (payload) => {
...
               if (api.assert(!(name in nftInstance.properties), 'cannot edit read-only properties')) {
                 nftInstance.properties[name] = data;
                 shouldUpdate = true;
               }
             } else {
+              oldProperty = nftInstance.properties[name];
               nftInstance.properties[name] = data;
               shouldUpdate = true;
             }
+            if (shouldUpdate) {
```
bt-cryptomancer, line 962: One optimization here would be to call the handleNftSetProperty action only if the data property value has actually changed.  If the new value is same as old value, no need to call this.  It's a small thing, but even loading & calling the other contract has some processing cost and saving every little bit helps.

bt-cryptomancer (changes requested):
Only a couple small suggestions for improvement (see comments on specific lines of code).  Also, there's an error with one of the nft contract test cases (the "does not set data properties" test), so please fix that.

image

eonwarped:
changes made. PTAL

bt-cryptomancer (approved):
Changes look good.

PR #86, opened 2021-08-01
Allow distribution contract to receive from DTF

donchate:


bt-cryptomancer (changes requested):
Can you clarify the purpose of adding the contractPayload parameter to the payRecipient function?  It looks like this new parameter is not actually used in any function calls so not sure why it's there.

donchate:
> Can you clarify the purpose of adding the contractPayload parameter to the payRecipient function? It looks like this new parameter is not actually used in any function calls so not sure why it's there.

I had a few work in progress branches and this got merged in here. The distribution itself also can pay out to a contract in the current code, but it wasn't complete. In order for it to work properly, it needs to do the same kind of integration action call like receiveDtfTokens, so this is lined up for the receiveDistTokens action.

Basically its just missing passing the contractPayload from the distribution to payRecipient. I will add that for completion sake of this PR.

bt-cryptomancer (approved):
Changes look good.

PR #87, opened 2021-08-02
addLiquidity payload to control maximum deviation from HE market

donchate:
When adding liquidity to a new pool, this allows user to set their own deviation from the HE market or disable this check entirely.


bt-cryptomancer (approved):
Changes look good.

PR #88, opened 2021-08-04
RPC error handling

donchate:


bt-cryptomancer (approved):
Changes look good, thanks for making this code safer!

PR #89, opened 2021-08-07
hotfix: distribution contract

donchate:
@bt-cryptomancer 
Used a return in validation when it should have been a continue / if block.

- Refactored all returns in the runDistribution method.
- Test case to confirm functionality.

*Need to sync contract and distro balances before deploying update.


bt-cryptomancer (approved):
Changes look good to me.

bt-cryptomancer (approved):
looks good

PR #90, opened 2021-08-11
Possible utility/governance token DTF creation from inflation contract

donchate:
A method to create and supply BEE:WORKERBEE and BEE:BEE token funds.

* test coverage for the inflation contract parts are in the tokenfunds test suite
* activation block number TBD
* requires #86 for distribution to accept funds


bt-cryptomancer (approved):
Changes look good.  Will fix test case after merge and final activation block number is set.

PR #93, opened 2021-08-26
utility/governance token creation fix

donchate:
Inflation contract runs every 1200 blocks, please set deployment block to a multiple of 1200 for successful deployment. Also there was a precision mismatch in PR #90 that is fixed here.


bt-cryptomancer (approved):
Changes look good.

PR #95, opened 2021-08-30
Throttle by vote count in addition to max posts processed

eonwarped:
Provides a more stable token maintenance mechanism, using the rewardInterval for how often to increment rewards and do post payout computation.

Moreover, within the same reward interval, we will process all posts within this interval together, possibly spread over multiple maintenance periods

In order to schedule multiple tokens fairly, we change the token maintenance order by looping and iterating by ID instead of by processing timestamp.

### DIFF
--- a/contracts/comments.js
+++ b/contracts/comments.js
@@ -305,7 +305,10 @@ async function computePostRewards(params, rewardPool, token, endTimestamp) {
   }
 }
 
-async function postClaimsInInterval(rewardPool, start, end) {
+async function postClaimsInInterval(params, rewardPool, start, end) {
+  const {
+    maxPostsProcessedPerRound,
+  } = params;
   let postOffset = 0;
   let newPendingClaims = api.BigNumber(0);
   let postsToPayout = await api.db.find('posts',
@@ -313,7 +316,7 @@ async function postClaimsInInterval(rewardPool, start, end) {
       rewardPoolId: rewardPool._id,
       cashoutTime: { $gte: start, $lte: end },
     },
-    1000,
+    maxPostsProcessedPerRound,
     postOffset,
     [{ index: 'byCashoutTime', descending: false }]);
   while (postsToPayout && postsToPayout.length > 0) {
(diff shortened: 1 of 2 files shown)


contracts/comments.js
```diff
@@ -212,105 +240,180 @@ async function payOutPost(rewardPool, token, post, timestamp, params) {
...
+  let postOffset = 0;
+  let newPendingClaims = api.BigNumber(0);
+  let postsToPayout = await api.db.find('posts',
+    {
+      rewardPoolId: rewardPool._id,
+      cashoutTime: { $gte: start, $lte: end },
+    },
+    maxPostsProcessedPerRound,
+    postOffset,
+    [{ index: 'byCashoutTime', descending: false }]);
```
bt-cryptomancer, line 325: Is it possible for two records to have the same byCashoutTime?  I think it would be good to add _id as a fallback index to sort on, as in:  ``[{ index: 'byCashoutTime', descending: false }, { index: '_id', descending: false }]``

contracts/comments.js
```diff
@@ -212,105 +240,180 @@ async function payOutPost(rewardPool, token, post, timestamp, params) {
...
+    }
+    postOffset += maxPostsProcessedPerRound;
+    postsToPayout = await api.db.find('posts',
+      {
+        rewardPoolId: rewardPool._id,
+        cashoutTime: { $gte: start, $lte: end },
+      },
+      maxPostsProcessedPerRound,
+      postOffset,
+      [{ index: 'byCashoutTime', descending: false }]);
```
bt-cryptomancer, line 343: Is it possible for two records to have the same byCashoutTime?  I think it would be good to add _id as a fallback index to sort on, as in:  ``[{ index: 'byCashoutTime', descending: false }, { index: '_id', descending: false }]``

bt-cryptomancer (changes requested):
Looks good, aside from the couple DB queries I commented on.

Why did you cancel the automated build?  Is there some problem with it?  I want to make sure you've linted this and the test cases pass.

Have you been running this on a test node with no issues?

Also, when would be a good time to schedule the release?  I'm a bit concerned about disrupting ongoing payouts if we release while in the middle of one.

contracts/comments.js
eonwarped: yes, it's possible. will add an index

contracts/comments.js
eonwarped: same

eonwarped:
> Looks good, aside from the couple DB queries I commented on.
> 
> Why did you cancel the automated build? Is there some problem with it? I want to make sure you've linted this and the test cases pass.
I sent multiple so wanted to cancel, ended up cancelling the wrong one, then decided to wait until a bit later to retrigger (like right now).
> 
> Have you been running this on a test node with no issues?
It's now on rishi's testnet, so far so good. I'm comparing contract event emittances too, and they are so far looking ok as well.
> 
> Also, when would be a good time to schedule the release? I'm a bit concerned about disrupting ongoing payouts if we release while in the middle of one.

There is no concern with timing, as basically it should not affect payout amounts, just the timing for when it happens

bt-cryptomancer (approved):
Changes look good, thanks!

PR #96, opened 2021-08-31
fix updateProposalApprovals weight calculation

donchate:
- The call from tokens contract was adding and updating unnecessary tokens to the accounts table. It will now only store and update voteTokens used by the account in various DTFs.
- Proposal weight now checks the token against the voteToken and updates only when needed.
- CreateSSC update to recalculate all weights and remove unnecessary tokens from accounts


bt-cryptomancer (approved):
Changes look good.

PR #97, opened 2021-09-03
pools - separate slippage and price impact, add minOut and maxIn

donchate:
- Slippage protection for `swapTokens` using optional parameters `minAmountOut` / `maxAmountIn` depending on trade direction. Transactions without these parameters or those that specify `maxSlippage` can still be executed. `maxSlippage` in the swapTokens action no longer has any effect.
- Price impact control is retained for adding liquidity. It is no longer applied to `swapTokens`.


bt-cryptomancer (approved):
Changes look good.

PR #99, opened 2021-09-06
Distribution: bonus curve adjustment fix

donchate:
- The adjustment to `timeFactor` when liquidity is added was not working as intended. It was resetting to current block time in essentially all cases. The changes in this PR correct this.
- Tests for these operations are in the `distribution` contract test suite.

Intended functionality:
Example: a position is increased by 100% (doubled) on period 4 after initial deposit; timeFactor will be increased by 2 periods / bonus level will be reduced by 2 periods. After adding liquidity the effective bonus shares remains consistent.

The 24 hour holding time remains in effect to prevent abuse, and with this correction to timeFactor, should only affect a small number of situations where liquidity is added and then increased by a great order of magnitude shortly after.

### DIFF
--- a/contracts/marketpools.js
+++ b/contracts/marketpools.js
@@ -362,11 +362,13 @@ actions.addLiquidity = async (payload) => {
     if (lp) {
       const existingShares = lp.shares;
       const finalShares = api.BigNumber(lp.shares).plus(newShares);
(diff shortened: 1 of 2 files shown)


bt-cryptomancer (approved):
Changes look good.

PR #100, opened 2021-09-07
Fix issue with reward pool not issuing due to precision mismatch

eonwarped:
Right now, it will fail to give rewards for any token whose precision is not 8.

Fixes where we get the symbol from.


bt-cryptomancer (approved):
Changes look good.

PR #102, opened 2021-09-21
dMarket Smart Contract

ali-h:
# dMarket Smart Contract
Dynamic Market Smart Contract allows users to trade tokens with different pairs other than SWAP.HIVE (the original market contract). A global pair means every token is allowed to be traded for it, only Hive Engine team can create a global pair. There are two global pairs allowed by default (SWAP.BTC & BEE). More pairs can be added by users for 100 BEE per token in a pair. It allows tribes and other projects to allow tokens to be traded for their native tokens (ex. DEC:PAL, ORB:PAL, BEE:PAL, SWAP.DOGE:PAL, etc..).

Transactions are similar to the original market contract with the addition of a "pair" parameter. 

Example: To buy BEE tokens for SWAP.BTC
```JSON
{
    "contractName": "dmarket",
    "contractAction": "buy",
    "contractPayload": {
        "symbol": "BEE",
        "pair": "SWAP.BTC",
        "quantity": "150",
        "price": "0.00000912"
    }
}
```

Complete Documentation: [pending]

### DIFF
--- a/contracts/dmarket.js
+++ b/contracts/dmarket.js
@@ -204,7 +204,7 @@ const updateVolumeMetric = async (symbol, pair, pairPrecision, quantity, add = t
   await api.db.update('metrics', metric);
 };
 
-const updatePriceMetrics = async (symbol, pair, price) => {
+const updatePriceMetrics = async (symbol, pair, pairPrecision, price) => {
   const blockDate = new Date(`${api.hiveBlockTimestamp}.000Z`);
   const timestampSec = blockDate.getTime() / 1000;
 
@@ -270,11 +270,21 @@ const updateAskMetric = async (symbol, pair) => {
   await api.db.update('metrics', metric);
 };
 
-const updateTradesHistory = async (type, buyer, seller, symbol, pair, pairPrecision, quantity, price, volume, buyTxId, sellTxId) => {
+const updateTradesHistory = async (type,
+  buyer,
+  seller,
+  symbol,
+  pair,
+  pairPrecision,
+  quantity,
+  price,
+  volume,
+  buyTxId,
+  sellTxId) => {
   const blockDate = new Date(`${api.hiveBlockTimestamp}.000Z`);
   const timestampSec = blockDate.getTime() / 1000;
   const timestampMinus24hrs = blockDate.setUTCDate(blockDate.getUTCDate() - 1) / 1000;
-  
+
   // clean history
   let tradesToDelete = await api.db.find(
     'tradesHistory',
@@ -321,7 +331,7 @@ const updateTradesHistory = async (type, buyer, seller, symbol, pair, pairPrecis
   newTrade.buyTxId = buyTxId;
   newTrade.sellTxId = sellTxId;
   await api.db.insert('tradesHistory', newTrade);
-  await updatePriceMetrics(symbol, pair, price);
+  await updatePriceMetrics(symbol, pair, pairPrecision, price);
 };
 
 const removeExpiredOrders = async (table) => {
@@ -392,7 +402,7 @@ const findMatchingSellOrders = async (order, tokenPrecision, pairPrecision) => {
     pair,
     priceDec,
   } = order;
-  
+
   const buyOrder = order;
   let offset = 0;
   let volumeTraded = 0;
@@ -452,7 +462,7 @@ const findMatchingSellOrders = async (order, tokenPrecision, pairPrecision) => {
             await api.db.update('sellBook', sellOrder);
           } else {
             if (api.BigNumber(qtyLeftSellOrder).gt(0)) {
-              // transfer the leftover tokens to the seller since the order can no longer be filled
+              // transfer remaining tokens to seller since the order can no longer be filled
               await api.transferTokens(sellOrder.account, symbol, qtyLeftSellOrder, 'user');
             }
             api.emit('orderClosed', { account: sellOrder.account, type: 'sell', txId: sellOrder.txId });
@@ -558,7 +568,7 @@ const findMatchingSellOrders = async (order, tokenPrecision, pairPrecision) => {
   if (api.BigNumber(buyOrder.quantity).gt(0)) {
     await api.db.update('buyBook', buyOrder);
   }
-  
+
   // update metrics
   if (api.BigNumber(volumeTraded).gt(0)) {
     await updateVolumeMetric(symbol, pair, pairPrecision, volumeTraded);
@@ -639,7 +649,7 @@ const findMatchingBuyOrders = async (order, tokenPrecision, pairPrecision) => {
             await api.db.update('buyBook', buyOrder);
           } else {
             if (api.BigNumber(buyOrdertokensLocked).gt(0)) {
-              // transfer the leftover tokens to the buyer since the order can no longer be filled
+              // transfer remaining tokens to buyer since the order can no longer be filled
               await api.transferTokens(buyOrder.account, pair, buyOrdertokensLocked, 'user');
             }
             api.emit('orderClosed', { account: buyOrder.account, type: 'buy', txId: buyOrder.txId });
@@ -777,14 +787,16 @@ actions.buy = async (payload) => {
 
   const token = await api.db.findOneInTable('tokens', 'tokens', { symbol });
 
+  // perform a few verifications
   if (!api.assert(token, 'symbol does not exist')) return;
-  if (!api.assert(countDecimals(quantity) <= token.precision, 'invalid quantity')) return;
+  if (!api.assert(countDecimals(quantity) <= token.precision
+    && api.BigNumber(quantity).gt(0), 'invalid quantity')) return;
 
   const pairInDb = await api.db.findOne('pairs', { pair });
 
   // check if symbol is included in allowedSymbols or the pair is global
   if (!api.assert(pairInDb && (pairInDb.allowedSymbols.indexOf(symbol) !== -1 || pairInDb.allowedSymbols === true), 'pair does not exist')) return;
-  if (!api.assert(api.BigNumber(price).gt(0) && countDecimals(price) <= pair.precision, 'invalid price')) return;
+  if (!api.assert(api.BigNumber(price).gt(0) && countDecimals(price) <= pairInDb.precision, 'invalid price')) return;
 
   const nbTokensToLock = api.BigNumber(price)
     .multipliedBy(quantity)
@@ -797,12 +809,7 @@ actions.buy = async (payload) => {
     });
 
     // make sure tokens are locked
-    if (api.assert(transferIsSuccessful(tokenTransfer,
-      'transferToContract',
-      finalAccount,
-      CONTRACT_NAME,
-      pair,
-      nbTokensToLock), 'failed to transfer tokens')) {
+    if (api.assert(transferIsSuccessful(tokenTransfer, 'transferToContract', finalAccount, CONTRACT_NAME, pair, nbTokensToLock), 'failed to transfer tokens')) {
       const blockDate = new Date(`${api.hiveBlockTimestamp}.000Z`);
       const timestampSec = blockDate.getTime() / 1000;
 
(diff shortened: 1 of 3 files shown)


Rishi556:
> only Hive Engine team can create a global pair

Why not let the witnesses decide? If a majority of top witnesses(11/20 right now) are voting for a particular token to be global, then it should be.

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+// fee in UTILITY TOKEN required to create a new pair
+const PAIR_CREATION_FEE = '500';
+
+const CONTRACT_NAME = 'dmarket';
+
+actions.createSSC = async () => {
+  const tableExists = await api.db.tableExists('pairs');
+
+  if (tableExists === false) {
+    await api.db.createTable('pairs', ['symbol', 'precision', 'allowedSymbols']);
```
eonwarped, line 17: Please reduce the number of indexes, as they take up a lot of space. We should audit which queries you are using and optimize for that. Note that from recent times, we now have the ability to do compound indexes (meaning if you always are querying with pair/symbol while sorting by e.g. price, then you can make a compound index on [pair, symbol, priceDec]. For example. See https://github.com/hive-engine/steemsmartcontracts/blob/hive-engine/contracts/comments.js#L18 for examples

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+    await api.db.createTable('buyBook', ['symbol', 'pair', 'account', 'priceDec', 'expiration', 'txId']);
+    await api.db.createTable('sellBook', ['symbol', 'pair', 'account', 'priceDec', 'expiration', 'txId']);
+    await api.db.createTable('tradesHistory', ['symbol', 'pair']);
+    await api.db.createTable('metrics', ['symbol', 'pair']);
+
+    // default global pairs
+    const pairs = [
```
eonwarped, line 24: Should these be called "baseTokens"? Pair generally refers to two tokens, e.g. https://coinmarketcap.com/alexandria/glossary/pair

ali-h:
> > only Hive Engine team can create a global pair
> 
> Why not let the witnesses decide? If a majority of top witnesses(11/20 right now) are voting for a particular token to be global, then it should be.
> 
> I say this because right now we are too dependent on the engine team and I'd like to slowly start transitioning away from that wherever possible.

Yes, this is absolutely better fit. not very familiar with the witness contract yet, will have a peak. Perhaps a proposal kind-of system requiring vote between top 20 witnesses.

contracts/dmarket.js
ali-h: Yes, this does make more sense. Not good with naming things.

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+      {
+        pair: 'BEE',
+        precision: 8,
+        allowedSymbols: true, // true - all tokens are allowed
+      },
+      {
+        pair: 'SWAP.BTC',
```
eonwarped, line 31: I think we should use terminology consistent with https://www.investopedia.com/terms/c/currencypair.asp . In the terminology, price represents how much of the quote token is needed to purchase 1 unit of the base token. 

So that means this 'pair' should be renamed 'quoteToken'

Also suggest renaming 'allowedSymbols' to  'allowedBaseTokens'

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+        precision: 8,
+        allowedSymbols: true, // true - all tokens are allowed
```
eonwarped, line 33: Suggestion here is instead of re-using allowedSymbols, we have a separate indicator 'global' to check if present and use that instead.

eonwarped: In fact, maybe we should change the action to be 'setGlobalPair', so that it can be toggled on (but probably should not be allowed to turn it off)

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+          // update the volume
+          volumeTraded = api.BigNumber(volumeTraded).plus(qtyTokensToSend);
+
+          // set quantity to zero to stop the loop
+          buyOrder.quantity = '0';
+          await api.db.remove('buyBook', buyOrder);
+          api.emit('orderClosed', { account: buyOrder.account, type: 'buy', txId: buyOrder.txId });
+        }
+      } else {
+        let qtyTokensToSend = api.BigNumber(sellOrder.price)
```
eonwarped, line 494: I would like to simplify this loop. This else clause has a lot in common with the previous. Can we consolidate the common logic?

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+  ]);
+
+  do {
+    const nbOrders = buyOrderBook.length;
+    let inc = 0;
+
+    while (inc < nbOrders && api.BigNumber(sellOrder.quantity).gt(0)) {
+      const buyOrder = buyOrderBook[inc];
+      if (api.BigNumber(sellOrder.quantity).lte(buyOrder.quantity)) {
+        let qtyTokensToSend = api.BigNumber(buyOrder.price)
```
eonwarped, line 614: Same, always round down

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+          // update the volume
+          volumeTraded = api.BigNumber(volumeTraded).plus(qtyTokensToSend);
+
+          // set quantity to zero to stop the loop
+          sellOrder.quantity = 0;
+          await api.db.remove('sellBook', sellOrder);
+          api.emit('orderClosed', { account: sellOrder.account, type: 'sell', txId: sellOrder.txId });
+        }
+      } else {
+        let qtyTokensToSend = api.BigNumber(buyOrder.price)
```
eonwarped, line 671: Same comment to simplify loop

test/dmarket.js
```diff
@@ -0,0 +1,1144 @@
+
+/* eslint-disable */
+/* eslint-disable no-await-in-loop */
+/* eslint-disable no-undef */
+/* eslint-disable no-console */
+/* eslint-disable func-names */
+
+const { fork } = require('child_process');
```
eonwarped, line 8: Isn't this copied from the market test? Should be using the new fixtures. How easy would it be to adjust?

contracts/dmarket.js
ali-h: added a separate indicator 'isGlobal'

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+          // transfer the tokens to the seller
+          await api.transferTokens(sellOrder.account, pair, qtyTokensToSend, 'user');
+
+          // update the sell order
+          const qtyLeftSellOrder = api.BigNumber(sellOrder.quantity)
+            .minus(buyOrder.quantity)
+            .toFixed(tokenPrecision);
+          const nbTokensToFillOrder = api.BigNumber(sellOrder.price)
+            .multipliedBy(qtyLeftSellOrder)
+            .toFixed(pairPrecision);
```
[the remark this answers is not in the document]
ali-h: did this throughout the whole contract

contracts/dmarket.js
```diff
@@ -0,0 +1,1285 @@
...
+
+          if (api.BigNumber(qtyLeftSellOrder).gt(0)
+            && (api.BigNumber(nbTokensToFillOrder).gte('0.00000001'))) {
```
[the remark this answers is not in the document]
ali-h: added a check so now this '0.00000001' is replaced with minimum value of the current quoteToken (taking precision into account).

contracts/dmarket.js
ali-h: I did have a look but it just gets more complicated since there are little key differences. Not sure if possible in a simpler way.

contracts/dmarket.js
```diff
@@ -0,0 +1,1258 @@
...
+const PAIR_CREATION_FEE = '500';
+
+const CONTRACT_NAME = 'dmarket';
+
+actions.createSSC = async () => {
+  const tableExists = await api.db.tableExists('quoteTokens');
+
+  if (tableExists === false) {
+    await api.db.createTable('quoteTokens', ['quoteToken']);
+    await api.db.createTable('buyBook', ['symbol', 'quoteToken', 'account', 'priceDec', 'expiration', 'txId']);
```
eonwarped, line 18: [{ name: 'byPriceDecDesc', index: { 'quoteToken': 1, 'symbol': 1, 'priceDec': -1 }}, 'expiration']  appear to be the only indexes needed.

contracts/dmarket.js
```diff
@@ -0,0 +1,1258 @@
...
+      await api.db.remove(table, order);
+
+      if (table === 'buyBook') {
+        api.emit('orderExpired', { type: 'buy', txId: order.txId });
+
+        await updateAskMetric(order.symbol);
+      } else {
+        api.emit('orderExpired', { type: 'sell', txId: order.txId });
+
+        await updateBidMetric(order.symbol);
```
eonwarped, line 384: updateBidMetric, updateAskMetric appear to need a quoteToken as parameter. (Test coverage to catch this?)

contracts/dmarket.js
eonwarped: I do not see any queries by txId, but please confirm.

eonwarped: [{ name: 'byPriceDecAsc', index: { 'quoteToken': 1, 'symbol': 1, 'priceDec': 1}}, 'expiration']

contracts/dmarket.js
```diff
@@ -0,0 +1,1258 @@
...
+
+actions.createSSC = async () => {
+  const tableExists = await api.db.tableExists('quoteTokens');
+
+  if (tableExists === false) {
+    await api.db.createTable('quoteTokens', ['quoteToken']);
+    await api.db.createTable('buyBook', ['symbol', 'quoteToken', 'account', 'priceDec', 'expiration', 'txId']);
+    await api.db.createTable('sellBook', ['symbol', 'quoteToken', 'account', 'priceDec', 'expiration', 'txId']);
+    await api.db.createTable('tradesHistory', ['symbol', 'quoteToken']);
+    await api.db.createTable('metrics', ['symbol', 'quoteToken']);
```
eonwarped, line 21: remove indexes (leave as []), and add primaryKey, e.g.:

await api.db.createTable('metrics', [], { primaryKey: ['quoteToken', 'symbol'] });

contracts/dmarket.js
```diff
@@ -0,0 +1,1258 @@
...
+const CONTRACT_NAME = 'dmarket';
+
+actions.createSSC = async () => {
+  const tableExists = await api.db.tableExists('quoteTokens');
+
+  if (tableExists === false) {
+    await api.db.createTable('quoteTokens', ['quoteToken']);
+    await api.db.createTable('buyBook', ['symbol', 'quoteToken', 'account', 'priceDec', 'expiration', 'txId']);
+    await api.db.createTable('sellBook', ['symbol', 'quoteToken', 'account', 'priceDec', 'expiration', 'txId']);
+    await api.db.createTable('tradesHistory', ['symbol', 'quoteToken']);
```
eonwarped, line 20: await api.db.createTable('tradesHistory', [{ name: 'byTimestamp', index: { 'quoteToken': 1, 'symbol': 1, 'timestamp': 1}}]);

contracts/dmarket.js
```diff
@@ -0,0 +1,1258 @@
...
+  ]);
+
+  do {
+    const nbOrders = sellOrderBook.length;
+    let inc = 0;
+
+    while (inc < nbOrders && api.BigNumber(buyOrder.quantity).gt(0)) {
+      const sellOrder = sellOrderBook[inc];
+      // gets the minimum quantity of quoteToken required to fill the order
+      const minimumToFillOrder = api.BigNumber(1).shiftedBy(-qtPrecision);
```
eonwarped, line 435: Is there a test that confirms this is the right quantity?

contracts/dmarket.js
ali-h: updated, also added test units for expired orders

PR #106, opened 2021-10-22
only do maintenance on null ops

eonwarped:
This will fix issues with maintenance actions failing maintenance due to not being able to issue tokens


bt-cryptomancer (approved):
Changes look good to me.

PR #115, opened 2021-10-29
Rollforward: Keep track of previously processed posts

eonwarped:
Rollforward of #111  with fix.

Fixes interval adding all posts within range, and allows comments to paid out posts to be properly indexed, and does not allow comments to re-activate after payout.

Adds new table called 'postMetadata' that maps authorperm to rewardPoolIds. Can be extended to store more in the future if needed. Lookups are by primary key _id, should be efficient, and does not affect the index for posts for a given reward pool.

This also adds a node change to make it so that MongoErrors properly error out completely rather than allow the node to infinite loop.

### DIFF
--- a/contracts/comments.js
+++ b/contracts/comments.js
@@ -5,6 +5,7 @@
 const SMT_PRECISION = 10;
 const MAX_VOTING_POWER = 10000;
 const MAX_WEIGHT = 10000;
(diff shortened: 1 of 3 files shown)


bt-cryptomancer (approved):
Changes look good.

PR #117, opened 2021-11-09
Comments reset pool

eonwarped:
Allows token issuer to reset reward pool, which will set claims to 0, reward pool to 0, and reset start time so that new votes will accumulate claims as well.


bt-cryptomancer (approved):
Changes look good.

PR #119, opened 2021-11-22
Fix handle nft change with block number

eonwarped:
Allows replay from previous snapshot to gate new behavior on block number.


bt-cryptomancer (approved):
Looks good to me.

PR #121, opened 2021-11-24
Nft miner fix handle nft change

eonwarped:
Restore update functionality, and skip inactive pools .

Note that if a pool is inactive, to reactivate the pool owner should change a setting to recompute mining power right before so any nft changes since deactivation can be computed again. This can be a mining setting or a burn change action.

To deploy: first deploy mining, followed by nft.


bt-cryptomancer (approved):
Changes look good to me.

PR #123, opened 2021-12-01
Crash debug logging using loglevel library

eonwarped:
Added config parameter to tweak log level to help in crash scenarios.


bt-cryptomancer (changes requested):
The actual code changes look good, however I noticed you updated the package-lock.json file to use lockfileVersion 2.  I believe this version 2 format requires a newer version of npm.  Introducing the loglevel library shouldn't change the minimum version of npm required to run the node, right? So to avoid affecting compatibility with older versions of npm, I'd prefer to have package-lock.json reverted back to the previous lockfileVersion 1 format.  Unless you feel strongly about having it be lockfileVersion 2 for some reason.

4Ykw:
Tested myself these for already 2 weeks at least. So far no issues.

eonwarped:
Yeah, I think it just automatically updated the lockfile version when I did npm install . Don't think it changes node version compatibility. I think it's fine to just have people use the newer version of npm.

bt-cryptomancer (approved):
Okay, I did a bit of reading about lockfileVersion 2 and noticed that it is backwards compatible with npm 6.  I guess it doesn't really matter which version we use, and the underlying Node.js compatibility doesn't change either.  So if you guys have no concerns about it, then I'm fine with it too.

Happy to approve this PR now.

PR #125, opened 2021-12-06
Options for disabling downvote, and for ignoring decline payout

eonwarped:
Disabling downvote means any downvotes will just get set to 0 rshares. Ignoring decline payout means the reward pool does not respect the comment option, and will auto enable payout.


bt-cryptomancer (approved):
Changes look good.

PR #126, opened 2021-12-06
add index to posts, use explicit _id order

eonwarped:


bt-cryptomancer (approved):
Good catch, we should definitely be using explicit indexes on any query where the order matters.

PR #127, opened 2021-12-08
Witness schedule fix order

eonwarped:
This can cause divergence in log if payouts happen in a different order than others.


bt-cryptomancer (approved):
Changes look good.

PR #128, opened 2021-12-13
fix issue where non-typemap equipped nfts do not reset balance on update

eonwarped:
When an equipped nft is not in the type map, when we recompute balances, it fails to reset the nftBalances field.

Eventually, we should not need to do this anymore as we started to track equipped nft's and types directly in the miningpower object. This will make the update a lot faster as well. Needs subsequent fix.


bt-cryptomancer (approved):
Change looks good to me.

PR #129, opened 2021-12-15
app tax and excludeTags implementation for comments contract

eonwarped:
This implements the app-based beneficiary options for comments contract. If the post is created with a designated app, then no beneficiary is applied, otherwise, it will auto apply a beneficiary, where the designated percent is taken from the authorPortion first, before applying the other beneficiaries from the comment options, if any.

Also implements excludeTags config which will will prevent posts using specified tags from being indexed.


bt-cryptomancer (changes requested):
This looks good to me, but can you add a test case for when the appTaxConfig percent is 100?  I want to be sure that corner case works.  Also, this being the sensitive contract that it is, I'd like to see it running with no problems on a testnet for a while before releasing it.  How much testing have you done so far?

eonwarped:
The additions don't influence the load properties of the contract, so fairly confident it is safe. however, i will put it on the testnet soon.

bt-cryptomancer (approved):
Looks good, thanks for adding the new test case.  Let's let this sit on the testnet for a week or so and then can release if that goes smoothly.

bt-cryptomancer (approved):
Changes look good.

PR #131, opened 2022-01-03
New Contract: Roles

donchate:
### Description
This contract provides a stake-based remuneration capability for a DAO to establish community roles.
Community members may then, in a decentralized fashion, vie for positions based on stake-weighted approval gained from the voting token stakers.

Candidates are ranked by approval weight in the configured voting token. A number of regular (main) ranked slots receive recurring remuneration, and a number of alternate (backup) ranked slots receive remuneration using a random rotation of remaining eligible candidates. Token balances are fully distributed on each tick, so funds must be provided using manual deposit, or other smart contract (i.e DTF, Distribution).

More in [docs](https://github.com/donchate/steemsmartcontracts-wiki/blob/master/Roles-Contract.md).

### PR Details
- Does hook into _tokens_ contract for stake weighted state updates
- Unit testing intended to cover integral actions / functions within the contract. Integration testing (DTF, Distribution funding) TBD in testnet. Hope to do this concurrently with PR review.

### DIFF
--- a/contracts/roles.js
+++ b/contracts/roles.js
@@ -453,7 +453,7 @@ actions.applyForRole = async (payload) => {
   }
 };
 
-actions.toggleApplyForRole = async (payload) => {
+actions.setApplyActive = async (payload) => {
   const {
     roleId,
     active,
@@ -461,8 +461,7 @@ actions.toggleApplyForRole = async (payload) => {
   } = payload;
 
   if (!api.assert(isSignedWithActiveKey === true, 'you must use a transaction signed with your active key')
-    && !api.assert(typeof roleId === 'string', 'invalid roleId')
-    && !api.assert(typeof active === 'string', 'invalid active')) {
+    && !api.assert(typeof roleId === 'string', 'invalid roleId')) {
     return;
   }
   const role = await api.db.findOne('roles', { _id: api.BigNumber(roleId).toNumber() });
@@ -471,7 +470,7 @@ actions.toggleApplyForRole = async (payload) => {
     && api.assert(existingApply, 'candidate does not exist for sender')) {
     existingApply.active = !!active;
     await api.db.update('candidates', existingApply);
-    api.emit('toggleApplyForRole', { roleId: role._id, account: existingApply.account, active });
+    api.emit('setApplyActive', { roleId: role._id, account: existingApply.account, active });
   }
 };
 
--- a/contracts/roles.js
+++ b/contracts/roles.js
@@ -749,16 +749,11 @@ async function checkPendingCandidates(inst, params) {
     .dividedBy(api.BigNumber(10).pow(voteTokenObj.precision));
   const random = api.random();
 
-  const tickTime = api.BigNumber(blockDate.getTime())
-    .minus(params.instanceTickHours * 3600 * 1000).toNumber();
   const pendingRoles = await api.db.find('roles',
     {
       instanceId: inst.id,
       active: true,
       'tokenBalances.0': { $exists: true },
-      lastTickTime: {
-        $lte: tickTime,
-      },
     },
     params.maxTxPerBlock,
     0,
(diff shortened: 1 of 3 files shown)


contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+/* eslint-disable no-await-in-loop */
+/* global actions, api */
+
+const ContractName = 'roles';
+const FeeMethod = ['burn', 'issuer'];
+
+actions.createSSC = async () => {
+  const tableExists = await api.db.tableExists('instances');
+  if (tableExists === false) {
+    await api.db.createTable('instances', ['id', 'lastTickTime']);
```
bt-cryptomancer, line 12: I think you can remove `'id'` from `['id', 'lastTickTime']`

If you were intending to add an index for the implicit MongoDB _id field (I'm assuming id without the underscore is a typo), there's always an index for that and you don't have to add it explicitly.

eonwarped: Yes, I think we should be careful here, I've seen parts using .id vs ._id. It does seem like we want to rely on the auto index _id for the instance table.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+      oldApprovalWeight = acct.weights[wIndex].weight;
+      acct.weights[wIndex].weight = approvalWeight;
+
+      const deltaApprovalWeight = api.BigNumber(approvalWeight)
+        .minus(oldApprovalWeight)
+        .dp(token.precision, api.BigNumber.ROUND_HALF_UP);
+
+      if (!api.BigNumber(deltaApprovalWeight).eq(0)) {
+        await api.db.update('accounts', acct);
+        const approvals = await api.db.find('approvals',
```
eonwarped, line 732: I think we should query by voteToken here, to limit the scope for a given token stake change

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+    params.maxRolesPerBlock,
+    0,
+    [{ index: 'byLastTickTime', descending: false }, { index: '_id', descending: false }]);
+
+  for (let i = 0; i < pendingRoles.length; i += 1) {
+    const role = pendingRoles[i];
+    const funded = [];
+    const payTokens = role.tokenBalances.filter(t => api.BigNumber(t.quantity).gt(0));
+    const totalSlots = api.BigNumber(role.mainSlots).plus(role.backupSlots).toNumber();
+    const roleTickTime = api.BigNumber(blockDate.getTime())
```
eonwarped, line 797: This probably does not need BigNumber

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+      .minus(role.tickHours * 3600 * 1000).toNumber();
+
+    if (role.lastTickTime <= roleTickTime) {
```
eonwarped, line 800: this could be part of the query as well? currently it checks <= instTickTime but we can have it be <= min(two tick times)

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+                  .times(random))
+                .toFixed(voteTokenObj.precision, api.BigNumber.ROUND_HALF_UP);
+            }
+
+            accWeight = api.BigNumber(accWeight)
+              .plus(candidate.approvalWeight.$numberDecimal)
+              .toFixed(voteTokenObj.precision, api.BigNumber.ROUND_HALF_UP);
+
+            if (candidate.active === true) {
+              if (funded.length < role.mainSlots || api.BigNumber(backupWeight).lte(accWeight)) {
```
eonwarped, line 832: The backup mechanism seems a bit weird, or works only when numBackup = 1. If there are more than 1 backups requested, and it generates a single number, it'll just pick two backups together without re-rolling the die?  I think this might be similar to what witness contract does (Which is also wrong)

test/roles.js
```diff
@@ -0,0 +1,942 @@
...
+        refHiveBlockNumber: refBlockNumber,
+        refHiveBlockId: 'ABCD1',
+        prevRefHiveBlockId: 'ABCD2',
+        timestamp: '2021-03-12T00:00:00',
+        transactions,
+      };
+
+      await fixture.sendBlock(block);
+      
+      // let res = await fixture.database.getLatestBlockInfo();
```
eonwarped, line 587: remove

test/roles.js
```diff
@@ -0,0 +1,942 @@
...
+      assert.ok(res.virtualTransactions.length == 0, 'Expected to not find virtualTransactions'); 
+
+      resolve();
+    })
+      .then(() => {
+        fixture.tearDown();
+        done();
+      });
+  });  
+
```
eonwarped, line 940: does this test deposits via dtf/distribution contracts?

test/roles.js
```diff
@@ -0,0 +1,942 @@
...
+    prevRefHiveBlockId: 'ABCD2',
+    timestamp: '2018-05-31T01:00:00',
+    transactions,
+  };
+
+  await fixture.sendBlock(block);
+  await tableAsserts.assertNoErrorInLastBlock();
+}
+
+// distribution test suite
```
eonwarped, line 138: roles test suite

eonwarped (changes requested):

test/roles.js
donchate: Not currently, I was going to confirm in testnet

contracts/roles.js
donchate: Makes sense since it was copied from witness contract. Going to rework.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+      const inst = await api.db.findOne('instances', { _id: role.instanceId });
+      if (inst.voteToken !== deltaToken.symbol) return true;
+    }
+    candidate.approvalWeight = {
+      $numberDecimal: api.BigNumber(candidate.approvalWeight.$numberDecimal)
+        .plus(deltaApprovalWeight),
+    };
+    await api.db.update('candidates', candidate);
+
+    const role = await api.db.findOne('roles', { _id: candidate.roleId });
```
bt-cryptomancer, line 114: if deltaToken is defined, then this is a redundant query (you've already looked up the role for deltaToken processing above).  It would be more efficient to re-use the earlier query results in that case, and only execute the query again here if deltaToken is not defined.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+        && typeof candidateFee.method === 'string'
+        && FeeMethod.indexOf(candidateFee.method) !== -1
+        && typeof candidateFee.symbol === 'string'
+        && typeof candidateFee.amount === 'string' && api.BigNumber(candidateFee.amount).gte(0), 'invalid candidateFee properties')) return;
+      const feeTokenObj = await api.db.findOneInTable('tokens', 'tokens', { symbol: candidateFee.symbol });
+      if (!api.assert(feeTokenObj && api.BigNumber(candidateFee.amount).dp() <= feeTokenObj.precision, 'invalid candidateFee token or precision')) return;
+    }
+
+    const voteTokenObj = await api.db.findOneInTable('tokens', 'tokens', { symbol: voteToken });
+    if (!api.assert(voteTokenObj && voteTokenObj.stakingEnabled, 'voteToken must have staking enabled')) return;
```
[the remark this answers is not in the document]
bt-cryptomancer: I don't think we should ever allow there to be a potentially unbounded number of operations happening when there's a staking update, there should always be some well-defined upper limit that we are comfortable with.  Otherwise there should be a mechanism to split update processing across multiple blocks, with intermediate state being stored between blocks.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+      active: false,
+      creator: api.sender,
+      lastTickTime: now.getTime(),
+    };
+    const insertedInst = await api.db.insert('instances', newInstance);
+
+    if (api.sender !== api.owner
+        && api.sender !== 'null'
+        && api.BigNumber(instanceCreationFee).gt(0)) {
+      await api.executeSmartContract('tokens', 'transfer', {
```
bt-cryptomancer, line 169: Although the possibility of this transfer failing is remote, it would be safe to transfer the fees before creating the new instance, and verify that the transfer succeeded.  See, for example, the technique used in the issue action of the nft contract.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+      existingInst.candidateFee = candidateFee;
+    }
+
+    await api.db.update('instances', existingInst);
+
+    // burn the token update fees
+    if (api.sender !== api.owner
+        && api.sender !== 'null'
+        && api.BigNumber(instanceUpdateFee).gt(0)) {
+      await api.executeSmartContract('tokens', 'transfer', {
```
bt-cryptomancer, line 218: Same as above, should handle the fees and verify success before updating the instance.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+        totalApprovalWeight: { $numberDecimal: '0' },
+      };
+      const insertedRole = await api.db.insert('roles', newRole);
+      insertedRoles.push({ instanceId: insertedRole.instanceId, roleId: insertedRole._id, name: insertedRole.name });
+    }
+
+    if (api.sender !== api.owner
+        && api.sender !== 'null'
+        && api.BigNumber(roleCreationFee).gt(0)) {
+      await api.executeSmartContract('tokens', 'transfer', {
```
bt-cryptomancer, line 277: Same as above, should handle the fees and verify success first.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+      existingRole.tickHours = tickHours;
+    }
+
+    await api.db.update('roles', existingRole);
+
+    // burn the token update fees
+    if (api.sender !== api.owner
+        && api.sender !== 'null'
+        && api.BigNumber(roleUpdateFee).gt(0)) {
+      await api.executeSmartContract('tokens', 'transfer', {
```
bt-cryptomancer, line 356: Same as above, should handle the fees and verify success first.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+      approvalWeight: { $numberDecimal: '0' },
+    };
+    const insertedId = await api.db.insert('candidates', newCandidate);
+
+    if (api.sender !== api.owner && inst.candidateFee) {
+      if (inst.candidateFee.method === 'burn') {
+        await api.executeSmartContract('tokens', 'transfer', {
+          to: 'null', symbol: inst.candidateFee.symbol, quantity: inst.candidateFee.amount,
+        });
+      } else if (inst.candidateFee.method === 'issuer') {
```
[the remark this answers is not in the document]
bt-cryptomancer: I like this idea, more flexibility is always good and it feels slightly less clunky from a usage perspective.  Also, as per my other comments fees should be handled & verified before doing the insert.

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+    existingApply.active = !!active;
+    await api.db.update('candidates', existingApply);
+    api.emit('setApplyActive', { roleId: role._id, account: existingApply.account, active });
+  }
+};
+
+// deposit
+
+async function updateTokenBalances(role, token, quantity) {
+  const upRole = role;
```
bt-cryptomancer, line 480: Why use this `upRole` variable here instead of just directly using the passed in `role` ?

contracts/roles.js
```diff
@@ -0,0 +1,917 @@
...
+    await updateTokenBalances(role, depToken, quantity);
+    api.emit('receiveDtfTokens', {
+      roleId: data.roleId,
+      symbol,
+      quantity,
+    });
+  }
+};
+
+actions.receiveDistTokens = async (payload) => {
```
bt-cryptomancer, line 551: The code for this function is almost exactly the same as for receiveDtfTokens.  Perhaps you could introduce a helper function used by both of these, to avoid so much code duplication?

test/roles.js
```diff
@@ -0,0 +1,942 @@
...
+      res = (await fixture.database.getLatestBlockInfo());
+      // console.log(res);
+      await tableAsserts.assertNoErrorInLastBlock();
+
+      await assertWeightConsistency(1, 'PRO');
+      await assertWeightConsistency(2, 'PRO');
+      await assertWeightConsistency(3, 'PRO');
+      await assertWeightConsistency(4, 'PRO');
+
+      assert.ok(res.virtualTransactions.length > 0, 'Expected to find virtualTransactions');
```
bt-cryptomancer, line 759: For this test case, how are the ticks set up to generate these virtual transactions?  Maybe I missed it, but I didn't see any code to configure the per-block ticking behavior.

bt-cryptomancer (changes requested):
Overall logical structure & organization looks fine, but the devil is in the details as they say.  Please see my comments at specific code lines (and respond to all of eonwarped's comments).

Also, I'd like you to add some clear release instructions to the PR in regards to how to configure / setup the per-block ticking behavior for virtual transactions.  I think I remember we moved all that to config so no code changes would be needed to the main node software, but I don't recall all the details and don't want to be trying to figure it out when I'm in the middle of releasing this.

PR #132, opened 2022-01-03
Contract Update: marketpools fee

donchate:
Addition of a **0.25% fee** for swapping tokens. The fee is immediately deposited to liquidity reserves and increases the proportional value of all liquidity provider shares. LPs realize the gain when removing liquidity.

### DIFF
--- a/contracts/marketpools.js
+++ b/contracts/marketpools.js
@@ -2,7 +2,6 @@
 /* global actions, api */
 
 const TradeType = ['exactInput', 'exactOutput'];
-const TradeFeeMul = api.BigNumber('0.9975');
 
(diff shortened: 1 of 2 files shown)


bt-cryptomancer (changes requested):
Can you make the fee percentage configurable in the params table?  That way we can easily change it later on if desired, without having to do a contract update.

reazuliqbal:
Hi, is it possible to emit the fee charged as transaction event? It would be very helpful while doing calculation for tax purpose.

donchate:
Moved fee to params and added fee to emit.

bt-cryptomancer (approved):
Changes look good.

PR #133, opened 2022-01-04
add default sort explicitly to prevent dependency on "natural" order

eonwarped:
Tested on local node, matches perfectly with current behavior, and confirmed fixed divegence source for other witnesses.

### DIFF
--- a/libs/Database.js
+++ b/libs/Database.js
@@ -608,9 +608,8 @@ class Database {
           if (ind.length > 0) {
             const tableIndexes = await indexInformation(tableData);
 
-            let sort;
+            const sort = [];
             if (ind.every(el => tableIndexes[`${el.index}_1`] !== undefined || el.index === '$loki' || el.index === '_id' || tableIndexes[el.index] !== undefined)) {
-              sort = [];
               ind.forEach((el) => {
                 if (tableIndexes[el.index] !== undefined) {
                   tableIndexes[el.index].forEach((indexPart) => {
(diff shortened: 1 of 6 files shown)


libs/Database.js
```diff
@@ -632,6 +631,9 @@ class Database {
               // not exist.
               console.log(`Index ${JSON.stringify(ind)} not available for ${finalTableName}`); // eslint-disable-line no-console
             }
+            if (sort.length === 0 || sort[sort.length - 1][0] !== '_id') {
```
bt-cryptomancer, line 634: I don't understand why this logic is included:  `sort[sort.length - 1][0] !== '_id'`

Can you clarify the intent?  If you are aiming to add _id to the sort list if it isn't already in there, why just check the last element instead of all of them?  What if _id is in position 0 instead of position `sort.length - 1` ?

bt-cryptomancer (commented):
Looks good aside from the one line of logic I commented on.  Will approve after that is clarified.

libs/Database.js
eonwarped: Will change to check all, for some reason I thought I needed it as a tiebreak even if specified earlier in the sort order but it's not needed

bt-cryptomancer: Thanks, looks good with the change.  Good idea to include it as a tiebreak, but there's no need for it to be potentially specified twice in the sort order.

bt-cryptomancer (approved):
Changes look good to me.

PR #134, opened 2022-01-17
Add extra validation for marketAccount and officialMarket settings

bt-cryptomancer:
This is to prevent an error that can happen if a buyer tries to set itself as the market account to receive fees from NFT purchases.  Here's an example of one such invalid transaction:

https://he.dtools.dev/tx/63002c925da6f3a585f531bec353d45359c84f69

The solution is to disallow cases where the buyer account is the same as the marketAccount parameter or the officialMarket setting of the market.


eonwarped (approved):
Looks good

PR #144, opened 2022-04-29
Hive-engine light node

primersion:
## Description

This pull requests adds light node functionality to hive-engine. Instead of storing all data of the hive-engine blockchain, light nodes only store smart contract data and data required for block verification, but not the full chain (blocks, transactions are cleaned up periodically). This will allow for faster restoring / synchronizing and make the network / the witness block signing process more stable.

## How to enable

Stop the node. Set ```lightNode: true``` in the ```config.json```. Start the node and it will initially delete all transactions and blocks ```< lastVerifiedBlockNumber - blocksToKeep``` from the database. This process may take a few minutes - if you want to be faster you may want to use a lightnode snapshot and restore the database first. Ask me on discord for a snapshot or check out the following url:
https://snap.primersion.com/light/

Restoring process is the same as for full nodes:
* Download snapshot
* Drop database
* Restore snapshot using mongorestore
* Set ```startHiveBlock``` in ```config.json```
* Set flag ```lightNode: true``` in ```config.json```
* Start node

If switching from a full node you may need to run ```db.runCommand({compact: 'chain', force: true})``` to free up storage consumed by mongodb.

## Differences of light nodes

* getStatus request will return an additional parameter ```lightNode: true``` and ```blocksToKeep: 14400```
* getBlockInfo rpc call will only work from block ```lastVerifiedBlockNumber - blocksToKeep``` up to the head block
* getTransactionInfo rpc call will only work from block ```lastVerifiedBlockNumber - blocksToKeep``` up to the head block
* find_divergent_block script only compares the hash of the head block to the head block hash of another node

## Cons

* Not all transactions are stored in the database (only txs of the last ```blocksToKeep``` blocks)
* Blocks are only stored as long as >= ```lastVerifiedBlockNumber - blocksToKeep```
* Can't switch back from light node to full node (need to restore from full node snapshot)

## Pros

* Fast restore and dump (5 minutes, compared to > 2 hours for full node)
* Small dump size (360MB currently, compared to 40GB for full node)
* Nodes get less load
* 2GB of storage compared to 90GB of a full node

A few more details in the post over here: https://peakd.com/hiveengine/@primersion/hive-engine-light-nodes

### DIFF
--- a/libs/Database.js
+++ b/libs/Database.js
@@ -1029,15 +1029,18 @@ class Database {
 
   /**
    * Used by light nodes to cleanup (unneeded) blocks already verified
-   * by witnesses <= @cleanupUntilBlock
-   * @param cleanupUntilBlock
+   * by witnesses <= lastVerifiedBlockNumber - blocksToKeep
    * @returns {Promise<void>}
    */
-  async cleanupBlocks(cleanupUntilBlock) {
+  async cleanupBlocks() {
     if (!this.lightNode) {
       return;
     }
-    await this.chain.deleteMany({ $and: [{ _id: { $gt: 0 } }, { _id: { $lte: cleanupUntilBlock - this.blocksToKeep } }] }, { session: this.session });
+    const params = await this.findOne({ contract: 'witnesses', table: 'params', query: {} });
+    if (params && params.lastVerifiedBlockNumber) {
+      const cleanupUntilBlock = params.lastVerifiedBlockNumber - 1 - this.blocksToKeep;
+      await this.chain.deleteMany({ $and: [{ _id: { $gt: 0 } }, { _id: { $lte: cleanupUntilBlock } }] }, { session: this.session });
+    }
   }
 
   /**
--- a/libs/Database.js
+++ b/libs/Database.js
@@ -71,6 +71,7 @@ class Database {
     this.session = null;
     this.contractCache = {};
     this.objectCache = {};
+    this.lightNode = false;
   }
 
   startSession() {
@@ -132,10 +133,11 @@ class Database {
     });
   }
 
-  async init(databaseURL, databaseName) {
+  async init(databaseURL, databaseName, lightNode = false) {
     // init the database
     this.client = await MongoClient.connect(databaseURL, { useNewUrlParser: true, useUnifiedTopology: true });
     this.database = await this.client.db(databaseName);
+    this.lightNode = lightNode;
     // await database.dropDatabase();
     // return
     // get the chain collection and init the chain if not done yet
@@ -218,6 +220,9 @@ class Database {
   }
 
   async getTransactionInfo(txid) {
+    if (this.lightNode) {
+      throw new Error('getTransactionInfo not available for light node');
+    }
     const transactionsTable = this.database.collection('transactions');
 
     const transaction = await transactionsTable.findOne({ _id: txid }, { session: this.session });
@@ -240,6 +245,10 @@ class Database {
     const finalBlock = block;
     finalBlock._id = await this.getNextSequence('chain'); // eslint-disable-line no-underscore-dangle
     await this.chain.insertOne(finalBlock, { session: this.session });
+    // do not insert transactions for lightnode
+    if (this.lightNode) {
+      return;
+    }
     await this.addTransactions(finalBlock);
   }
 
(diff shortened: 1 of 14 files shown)


Rishi556:
Less memory is also required to run, and apps which only need current data to run(ex just care about current NFT inventory etc) can self host a node for a lot cheaper reducing the load on public infastructure.

libs/Database.js
```diff
@@ -1015,6 +1024,50 @@ class Database {
...
+   * Used by light nodes to cleanup (unneeded) blocks already verified
+   * by witnesses <= @lastVerifiedBlockNumber
+   * @param lastVerifiedBlockNumber
+   * @returns {Promise<void>}
+   */
+  async cleanupBlocks(lastVerifiedBlockNumber) {
+    if (!this.lightNode) {
+      return;
+    }
+    // await this.chain.deleteMany({ _id: { $lte: lastVerifiedBlockNumber } }, { session: this.session });
```
bt-cryptomancer, line 1038: Can you remove this comment?  No need to leave commented code lying around, better to clean it up.

bt-cryptomancer (changes requested):
I'm not completely comfortable with exposing `cleanupBlocks` to the smart contracts API.  I also don't like the idea of modifying the witnesses contract to do that, because this is a breaking change which will force everyone to upgrade their node software.  I'd rather have the next version release be an optional upgrade, so if you don't want to run a light node you don't need to upgrade.

Can you remove that from the witnesses contract and have it be done internally to the node software?  It could be done as a maintenance operation every so often (maybe every 50 blocks, say) inside the `produceBlock` function of `libs/Block.js`.  Also, please make sure that running a light node does not affect database hashes.

Can you merge the changes from this PR by @Rishi556 (which I've already approved) into your own branch:  https://github.com/hive-engine/steemsmartcontracts/pull/145

That PR makes changes to the `getStatus` function, which may cause a conflict since you touch that function too.  Would be convenient to have that resolved in your PR so I don't have to do conflict resolution when preparing the next release.

eonwarped:
> I'm not completely comfortable with exposing `cleanupBlocks` to the smart contracts API. I also don't like the idea of modifying the witnesses contract to do that, because this is a breaking change which will force everyone to upgrade their node software. I'd rather have the next version release be an optional upgrade, so if you don't want to run a light node you don't need to upgrade.
> 
> Can you remove that from the witnesses contract and have it be done internally to the node software? It could be done as a maintenance operation every so often (maybe every 50 blocks, say) inside the `produceBlock` function of `libs/Block.js`. Also, please make sure that running a light node does not affect database hashes.
> 


Agree with this ask. Right now the witness contract changes wouldn't even work as is because it would need it to be deployed and all nodes would have to understand this action. But this isn't the right way to do the clean anyway, and @bt-cryptomancer suggestion is better to have it happen internally for light nodes in the appropriate location (produce block is sensible)

primersion:
Thanks for the review and for your comments regarding the changes. Running a light node does not affect database hashes. Everything is the same as for full nodes, except that light nodes do not store all blocks / transactions.

I have now updated the code as follows:
* Merged @Rishi556 PR #145 
* Removed the comment
* Moved the cleaning up of blocks / transactions to Block.js (done every 100 blocks)
* Added another param ```blocksToKeep```, which defines how many blocks (and the corresponding txs) should be kept before cleaning up (default is currently 14400 ~ 12h)

libs/Database.js
```diff
@@ -1026,47 +1021,51 @@ class Database {
...
 
-    await this.chain.deleteMany({ $and: [{ _id: { $gt: 0 } }, { _id: { $lte: lastVerifiedBlockNumber } }] }, { session: this.session });
+  /**
+   * Used by light nodes to cleanup (unneeded) blocks already verified
+   * by witnesses <= lastVerifiedBlockNumber - blocksToKeep
+   * @param cleanupUntilBlock cleanup blocks with a smaller blockNumber
+   * @returns {Promise<void>}
+   */
+  async cleanupBlocks(cleanupUntilBlock) {
+    await this.chain.deleteMany({ $and: [{ _id: { $gt: 0 } }, { _id: { $lte: cleanupUntilBlock } }] }, { session: this.session });
```
bt-cryptomancer, line 1048: Should this be >= 0 instead of > 0?  There is a block 0, the genesis block.

libs/Database.js
```diff
@@ -1026,47 +1021,51 @@ class Database {
...
    */
   async wasLightNodeBefore() {
-    const transactionsTable = this.database.collection('transactions');
-    const someTransaction = await transactionsTable.findOne({}, { session: this.session });
-    if (someTransaction) {
-      // transaction table should be empty for a light node
-      return false;
-    }
-    return true;
+    const transaction = await this.getTransactionInfo(0);
```
bt-cryptomancer, line 1067: Not sure if this will work as intended.  getTransactionInfo expects a string tx ID, and just looking at the first few entries in the transactions table I see stuff like this:

image

I think it makes more sense to use `this.getBlockInfo(0)` for this check.

plugins/JsonRPCServer.js
```diff
@@ -19,6 +19,54 @@ let serverRPC = null;
...
+
+      // gets the domain of the SSC node
+      result.domain = config.domain;
+
+      // get the ssc chain id from config
+      result.chainId = config.chainId;
+
+      // get light node config of the SSC node
+      result.lightNode = config.lightNode;
+      result.blocksToKeep = config.blocksToKeep;
```
bt-cryptomancer, line 55: I think blocksToKeep should only be added to this status info if lightNode = true (since otherwise it doesn't apply and so might be confusing if it's present).

bt-cryptomancer (changes requested):
Mostly looks good now, just a couple minor points that I'm concerned about (see comments below).

Also, can you put some comments in the code wherever it seems appropriate to indicate that blocksToKeep only applies in the event that lightNode is true?

libs/Database.js
primersion: It was done this way because `createGenesisBlock` in `Blockchain.js` will create the genesis block again if it's not there and syncing will start from block 1 again. So block 0 is currently kept also on light nodes. I can either add a comment why it was done this way or change the `createGenesisBlock` method to exclude light nodes. What do you like better?

libs/Database.js
primersion: Good point. But `this.getBlockInfo(0)` does not work either, as light nodes also store the genesis block (see comment above). If I update the code as suggested above, this would work - otherwise I would have to think of another solution.

libs/Database.js
bt-cryptomancer: Okay I see.  That's fine then, no problem with keeping the genesis block for light nodes.  You can just add an explanatory comment.

libs/Database.js
bt-cryptomancer: What about simply using `this.getBlockInfo(1)` ?

bt-cryptomancer (approved):
This all looks good now, thanks for the revisions.  Will include this PR in the next version release.

PR #145, opened 2022-05-03
Visiting / will get status instead directing to a cannot get / page

Rishi556:
Small quality of life improvement. Visiting just the node's url as of now will lead to the default express no route page(see https://api.hive-engine.com/rpc/) but with this change, it returns the stats from the getStats call(see https://engine.rishipanthee.com/).


bt-cryptomancer (approved):
This looks fine to me.  Will include in the next version release.

PR #146, opened 2022-05-18
Hive engine - Add NATed-ports customization feature for witness management

4Ykw:
# Motivation
Some infrastructures run multiple nodes behind the same IPv4, and the ability to customize port forwarding on witness management via the already available script (`witness_action.js`) allows the witness external ports to be registered/unregistered behind the firewall on a node configured to use those NATed ports.

# Compatibility
If no variables are used under the `.env` file, the `config.js` ports will be used as before. Otherwise, if specified in `.env` they will be used ONLY when registering/unregistering the witness (using the `witness_action.js` script). In this case, the user should have already configured the required port forwarding to the host that will use the registering command.

# Example
Node "A" (IP), using default RPC port 5000 and P2P port 5001, is running on a VM behind a Firewall.
The user wishes to use the external IP "B" to register the witness as public address with 15000 port for RPC, and 15001 port for P2P. Then, the following should be configured on the firewall that faces the internet:
- all TCP incoming traffic on "B" towards port 15000 should be forwarded to "A" over port 5000
- all TCP incoming traffic on "B" towards port 15001 should be forwarded to "A" over port 5001

# Functionality
Allow the user to run multiple Hive-Engine nodes behind the same IP using different public ports. It also provides additional functionality for test nodes or High Availability custom configs (using VMs or containerization), especially when using something like the recently proposed "light node" configuration by @primersion at https://github.com/hive-engine/steemsmartcontracts/pull/144 (but not limited to).

# Required configurations (for NATed ports usage)
- .env
  - RPCNODEPORT=<external port being forwarded to the RPC port defined at `config.js`>
  - P2PPORT=<external port being forwarded to the P2P port defined at `config.js`>

The `.env.example` was updated to reflect these changes.


bt-cryptomancer (approved):
Changes look good to me.

PR #147, opened 2022-06-28
Add IPv6 Support For Witnesses

Rishi556:
Currently witnesses don't have the ability to communicate with IPv6 witnesses due to malformed URL. 

```
2022-06-27 23:16:22 error: [P2P] Error posting to h-e / round 747286 / Error: connect ENETUNREACH 0.0.10.42:5001
2022-06-28 00:03:05 info: [P2P] http://2602:fb95:1::3:5001/p2p
```

The proper form is to add []'s around the v6 address to signify that it's a v6 address and not a port. This does that. @4Ykw found the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "steemsmartcontracts",
-  "version": "1.7.3",
+  "version": "1.7.4",
   "description": "",
   "main": "app.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


bt-cryptomancer (approved):
I'll approve this, however I'd recommend that Top 20 witnesses avoid the use of ipv6 for now if possible.  It's not currently known if various Engine systems that rely on witnesses (e.g. withdrawal / deposit gateways) are fully compatible with ipv6.  That's something we would need to carefully test.

For now, I will ensure that Engine systems do not rely on any witnesses that choose to use ipv6.

4Ykw:
I have also tested this on the light node (`http://[2407:7000:9cb7:b800:d9d3:d7c:db4b:a7cd]:5010`) and produced blocks without errors for some days already. Note: The version will still appear as `1.7.3` because I didn't change that part.

PR #152, opened 2022-07-20
Find divergent block retries

eonwarped:
Retry on failure, up to 3 times, and with 500ms back-off, to help with throttling issues on primary node.


bt-cryptomancer (approved):
Looks good to me.

PR #155, opened 2022-07-25
Add ability to verify transactions from witness_actions.js

Rishi556:
Appending -v to `witness_actions.js` will let user choose to verify their transaction, and with -e they can specify an engine node to verify against.


bt-cryptomancer (approved):
Changes look good.

PR #156, opened 2022-07-29
find divergent block - introduce -h flag to only compare head blocks

primersion:
* Appending -h to find_divergent_block.js will only compare head blocks instead of doing a high / low check
* Light nodes use the default check and start with `low = firstBlockNumber` instead of `low = 0` (`firstBlockNumber` is the first block stored by a light node)


bt-cryptomancer (approved):
Change looks good to me.

PR #157, opened 2022-08-05
Increase witness pay

bt-cryptomancer:
This PR doubles BEE inflation for witness pay from 100k BEE per year to 200k BEE per year, and also changes payouts to give staked tokens rather than liquid tokens.


eonwarped (approved):
Looks good

PR #158, opened 2022-08-05
Node restore / repair script

primersion:
Already includes PR #156 as it is required for the script to work for light nodes.

Script for helping to restore / repair hive-engine node databases. The restore mode will simply drop the
existing database and do a full restore.
The repair mode will delete all invalid blocks / transactions and drop all other collections and restore the
last valid state using a light node snapshot.

These are the different modes:
- FULL node:
  - Repair by executing `node restore_partial.js`     (~30-60 minutes)
  - Drop    by executing `node restore_partial.js -d -s https://snap.primersion.com/` (~6 hours)
- LIGHT node:
  - Drop    by executing `node restore_partial.js -d`  (~30-60 minutes)
  - Repair is not supported as dropping is faster

### DIFF
--- a/find_divergent_block.js
+++ b/find_divergent_block.js
@@ -12,9 +12,10 @@ const { Database } = require('./libs/Database');
 
 program
   .option('-n, --node [url]', 'compare with given node', 'https://api.hive-engine.com/rpc')
+  .option('-h, --head-only', 'compare only the head block')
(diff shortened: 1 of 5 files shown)


Rishi556 (approved):

bt-cryptomancer (approved):
Looks good to me.

PR #159, opened 2022-08-05
Run a divergence check before registering witnesses

Rishi556:
The scrip to find divergence will run before registering a witness. Can be skipped by using option -s. Uses head only comparison for speed.


Rishi556:
Also includes https://github.com/hive-engine/steemsmartcontracts/pull/156

bt-cryptomancer (approved):
Change looks good.

PR #160, opened 2022-08-06
Update Config.json

Rishi556:
Remove api.hive.blog from config as testing is done on it and has caused problems in the past. 
Change mongodb database url to 127.0.0.1 as for some reason localhost doesn't seem to work on ipv6 only systems for some reason whereas 127.0.0.1 works on both dual stacked and ipv6 only.


bt-cryptomancer (approved):
Change looks good.

PR #162, opened 2022-09-06
Allows for RPC Config

Rishi556:
Adds 3 new config parameters, `rpcConfig.maxLimit`, `rpcConfig.maxOffset` and `rpcConfig.logRequests`. `maxLimit` is the greatest limit that will be allowed in find queries, the highest possible value is 1000. `maxOffset` is the maximum allowable offset, with -1 being for no max. `logRequests` lets the requests to the node be logged, with the format of INCOMING_IP - INCOMING_BODY.

### DIFF
--- a/config.json
+++ b/config.json
@@ -22,6 +22,6 @@
     "rpcConfig" : {
         "maxLimit" : 1000,
         "maxOffset" : -1,
-        "logRequests" : true
+        "logRequests" : false
     }
 }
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -18,5 +18,10 @@
     "defaultLogLevel": "warn",
     "lightNode": false,
     "blocksToKeep": 864000,
-    "domain" : ""
+    "domain" : "",
+    "rpcConfig" : {
+        "maxLimit" : 1000,
+        "maxOffset" : -1,
+        "logRequests" : true
```
eonwarped, line 25: default false, otherwise I suspect this might be noisy.

eonwarped (dismissed):

eonwarped (approved):

